PR #3, opened 2021-01-13
Improve simple planner profiles to handle working frame

Levi-Armstrong:
The current implementation did not handle working frame information for cartesian waypoints correctly. The PR restructures the simple planner to have a single generator similar to the other planners and updates the different profiles to support handling of working frame.


tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_profile.h
```diff
@@ -42,26 +42,10 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
-                                                                  const ManipulatorInfo&)>;
-using CartCartStepGenerator = std::function<CompositeInstruction(const CartesianWaypoint&,
-                                                                 const CartesianWaypoint&,
-                                                                 const PlanInstruction&,
-                                                                 const PlannerRequest&,
-                                                                 const ManipulatorInfo&)>;
+using SimplePlannerStepGeneratorFn = std::function<CompositeInstruction(const PlanInstruction&,
+                                                                        const PlanInstruction&,
+                                                                        const PlannerRequest&,
+                                                                        const ManipulatorInfo&)>;
```
marip8, line 48: From what I understand, some other class needs to invoke this method to produce a `CompositeInstruction` that is the interpolated path. If that's correct, would it make more sense to create a base class with this method and give a pointer to that base class to the objects that need to use it to interpolate instructions? It seems like this PR adds several helper structs and functions (e.g. `LVSTransitionInfo`) that could be eliminated or better compartmentalized if you created implementations that do specific types of interpolation and contain their own implementation-specific parameters (e.g. longest valid segment length, min points)

Levi-Armstrong: @marip8 I moved everything from the step generator to the profile classes which simplified things. Let me know what you think.

tesseract_motion_planners/src/simple/profile/simple_planner_fixed_size_assign_plan_profile.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+  InstructionInfo info1(prev_instruction, request, global_manip_info);
+  InstructionInfo info2(base_instruction, request, global_manip_info);
+
+  Eigen::MatrixXd states;
+  if (!info1.has_cartesian_waypoint && !info2.has_cartesian_waypoint)
+  {
+    const Eigen::VectorXd& jp = info2.extractJointPosition();
+    if (info2.instruction.getPlanType() == PlanInstructionType::LINEAR)
+      states = jp.replicate(1, linear_steps_ + 1);
```
marip8, line 50: Is there a benefit to thinking about interpolation in terms of steps instead of number of states? My intuition would be to specify the number of states, so I always have to fight my confusion when I see `replicate(n_states + 1)`

tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_fixed_size_assign_plan_profile.h
```diff
@@ -0,0 +1,76 @@
...
+#include <tesseract_motion_planners/simple/profile/simple_planner_profile.h>
+#include <tesseract_motion_planners/simple/profile/simple_planner_utils.h>
+
+#ifdef SWIG
+%shared_ptr(tesseract_planning::SimplePlannerFixedSizeAssignPlanProfile)
+#endif  // SWIG
+
+namespace tesseract_planning
+{
+class SimplePlannerFixedSizeAssignPlanProfile : public SimplePlannerPlanProfile
```
marip8, line 45: You might consider making `SimplePlannerPlanProfile` a `struct` instead of a class and making the parameters of the implementation public members. Making getter/setter functions seems unnecessary

tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_fixed_size_assign_plan_profile.h
```diff
@@ -0,0 +1,76 @@
...
+
+  int getFreespaceSteps();
+  void setFreespaceSteps(int freespace_steps);
+
+  int getLinearSteps();
+  void setLinearSteps(int linear_steps);
+
+protected:
+  int freespace_steps_;
+  int linear_steps_;
```
marip8, line 71: Is it relevant to interpolate freespace or linear moves with a different number of states? I can't really think of a scenario where a single number of steps wouldn't suffice for 99% of our uses cases

tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_fixed_size_plan_profile.h
```diff
@@ -0,0 +1,84 @@
...
+
+  int getFreespaceSteps();
+  void setFreespaceSteps(int freespace_steps);
+
+  int getLinearSteps();
+  void setLinearSteps(int linear_steps);
+
+protected:
+  int freespace_steps_;
+  int linear_steps_;
```
marip8, line 69: Same comment about possibly using a single number of steps

tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_fixed_size_assign_plan_profile.h
Levi-Armstrong: Currently we use different values for each. I would agree that they are close if a configuration of the robots are the same but if it requires a configuration flip then the freespace would need to be higher. One example would be to assign this for the transitions and this single profile would be configured to handle both freespace and linear types where for freespace you would want to set higher to allow for configuration changes.

marip8: Would the user of this class actually know about a configuration change between the two input states ahead of time and be able to change this parameter? It seems like the LVS profile should used if you care about how far apart the interpolated waypoints should be, and that the fixed size profiles should be used when a user just needs to specify some arbitrary number of steps/states (in which case the distinction between freespace and linear moves doesn't really matter).

Levi-Armstrong: > Would the user of this class actually know about a configuration change between the two input states ahead of time and be able to change this parameter?

Yes and no, they have information about the process and would have some idea if it may be possible. In the case where you think it would be possible you would set the freespace to larger. As you mention the LVS is the default because it uses more information to try and determine what the ideal number of steps would be.

A profile is intended to handle the supported instruction types (FREESPACE and LINEAR) which an instruction can only be one of these which is why you have the ability to configure them individually.

tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_fixed_size_assign_plan_profile.h
Levi-Armstrong: I will remove the setters and getters and make the parameters public.

tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_fixed_size_plan_profile.h
Levi-Armstrong: Same as above making them public.

tesseract_motion_planners/src/simple/profile/simple_planner_fixed_size_assign_plan_profile.cpp
Levi-Armstrong: Not really, I have just gotten used to thinking in terms of steps because trajopt uses this terminology. Changeing things to use number of states versus steps would touch a lot of code which I would like to avoid at least for now. After the release we look at changing things.

tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_utils.h
```diff
@@ -0,0 +1,130 @@
...
+   */
+  Eigen::Isometry3d extractCartesianWorldPose() const;
+
+  /**
+   * @brief Extract the joint positiion from the instruction waypoint
+   * @details If the instruction does not have a joint/state waypoint this throws an exception
+   * @return Joint Position
+   */
+  const Eigen::VectorXd& extractJointPosition() const;
+};
```
marip8, line 82: This class technically works, but it seems like a convenient hack for helper functions that really belong in the `CartesianWaypoint` and `JointWaypoint` classes. Classes holding a const reference to some other object also makes me a little suspicious since it can't know about or control when that object goes out of scope. I don't know that you necessarily need to change this now, but I think the planners and other code that leverages waypoint definitions would benefit from these functions being a part of the waypoint implementations

Levi-Armstrong: This was intentional because the goal of the Command Language was to keep it independent of the other components. Just contain the minimal amount of information need to describe a motion request and not be dependent on other components of Tesseract. This purely meant to only be used internally and and is not intended to be used, if you want I can set the visibility to hidden so it is not in the public API?

marip8: I don't think changing the visibility of this class really solves the issue. I'm more concerned that other developers might accidentally use it improperly. I would also argue that the couple of functions about Cartesian local and world frames are critical for understanding/working with Cartesian waypoints. The waypoint itself doesn't need to hold on to IK/FK information, but I think it should be able to report it's location in either of those coordinate systems provided a map of transforms and an IK/FK solver.

We don't have to resolve this now, but I just wanted to bring this up as a point of discussion. I can open an issue or discussion to track this further if you want

marip8 (approved):

johnwason:
@Levi-Armstrong where did the equivalent of fixed_size_interpolation.h move to?

Levi-Armstrong:
https://github.com/ros-industrial-consortium/tesseract_planning/tree/master/tesseract_motion_planners/include/tesseract_motion_planners/simple/profile

It was renamed.

johnwason:
I don't see the utility functions like `fixedSizeInterpolateStateWaypoint` in the new API, and the function pointers in the plan profiles are also removed. What is the correct way to access this type of functionality now?

johnwason:
Ok please @ me when you make breaking changes, I am updating Python now but I didn't see that the API had changed before.

Levi-Armstrong:
Here is the PR which [cleaned](https://github.com/ros-industrial-consortium/tesseract_planning/pull/3/files) up the simple planners. The reason for the API change was the current interface did not have a way to support working_frame in the manipulator info which required things to change to fix the bug.

PR #17, opened 2021-01-19
Graph Taskflow Generator Revision

marip8:
This PR revises the graph Taskflow generator class to be easier to use.

## Assumptions

- End conditions
  - The required error task is only applicable to conditional nodes. I assign the error task as the connection at index 0 to all conditional tasks
  - The done callback is applicable to both conditional and non-conditional nodes.
    - If a conditional node has any edges, I do not connect that node to the done callback
    - If a conditional node is a leaf node (i.e. has no defined edges), the connection at index 0 is to the error task (as mentioned above) and the connection at index 1 is to the done task
    - If a non-conditional node is a leaf node, it's only connection is to the done callback
- No graph checking is implemented for loops and "island" states. I think we should allow users the freedom to define their graphs in whatever way possible. It's easy enough to dump the taskflow to a file and understand if the graph looks correct

All of these assumptions should be documented in the header file

## Test case

Below is the graph generated by the unit test using the revised graph Taskflow generator showing the assumptions detailed above

image

### DIFF
--- a/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
@@ -73,8 +73,9 @@ public:
    * as the output of a non-conditional tf::Task is void. If source is a conditional task, the order
    * of the destinations should correspond to the integer output of the conditional tf::Task **plus one**.
    * The connection at index zero is reserved for the error callback. As an example,if the output of the
-   * conditional tf::Task is 2, the taskflow would transition to the node identified by the ID at index 1 in the destinations input
-   * A leaf node is always connected to the done callback. When the leaf node is a conditional node, the done callback is the connection at index 1
+   * conditional tf::Task is 2, the taskflow would transition to the node identified by the ID at index 1 in the
+   * destinations input A leaf node is always connected to the done callback. When the leaf node is a conditional node,
+   * the done callback is the connection at index 1
    *
    * Remember: when source is a conditional node, the connection at index 0 is always to the error callback
    *
--- a/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
@@ -69,16 +69,12 @@ public:
 
   /**
    * @brief Adds directed edges from a source node to desintation nodes in the taskflow graph
-   * @details If source is a non-conditional task, it is only relevant to provide one destination
-   * as the output of a non-conditional tf::Task is void. If source is a conditional task, the order
-   * of the destinations should correspond to the integer output of the conditional tf::Task **plus one**.
-   * The connection at index zero is reserved for the error callback. As an example,if the output of the
-   * conditional tf::Task is 2, the taskflow would transition to the node identified by the ID at index 1 in the
-   * destinations input A leaf node is always connected to the done callback. When the leaf node is a conditional node,
-   * the done callback is the connection at index 1
-   *
-   * Remember: when source is a conditional node, the connection at index 0 is always to the error callback
-   *
+   * @details If source is a non-conditional task, it is only relevant to provide one destination as the output of a
+   * non-conditional tf::Task is void. If source is a conditional task, the order of the destinations should correspond
+   * to the integer output of the conditional tf::Task. For example,if the output of the conditional tf::Task is 0, the
+   * taskflow would transition to the node identified by the ID at index 0 in the destinations input. A leaf node (i.e.
+   * node without defined edges) is always connected to the done callback. If the leaf node is a conditional node, it is
+   * connected to the error task at index 0 and the done task at index 1
    * @param from The ID of the source node of the edge
    * @param dest A list of IDs of the destination nodes of the edges.
    */
--- a/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
@@ -84,7 +84,7 @@ private:
   /** @brief Helper struct for holding relevant information about the nodes */
   struct Node
   {
-    Node(TaskGenerator::UPtr process_, const bool is_conditional_);
+    Node(TaskGenerator::UPtr process_, bool is_conditional_);
 
     TaskGenerator::UPtr process;
     const bool is_conditional;
--- a/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
@@ -40,47 +40,13 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 namespace tesseract_planning
 {
 /**
- * @brief This class generates taskflow graph. It allows you to connect different channels from the source to the
- *        destination process
+ * @brief This class facilitates the composition of an arbitrary taskflow graph.
+ * Tasks are nodes in the graph connected to each other in a configurable order by directed edges
  */
 class GraphTaskflow : public TaskflowGenerator
 {
 public:
   using UPtr = std::unique_ptr<GraphTaskflow>;
-
-  enum class NodeType : int
-  {
-    TASK = 0,
-    CONDITIONAL = 1
-  };
-
-  enum class SourceChannel : int
-  {
-    NONE = 0,
-    ON_SUCCESS = 1,
-    ON_FAILURE = 2
-  };
-
-  enum class DestinationChannel : int
-  {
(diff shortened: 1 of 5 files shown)


mpowelson:
Does this allow you to still define 0 edges that don't go to the error callback? Sending 0 returns to error when no explicit edges are defined is fine, but I am currently sending failures of tasks to things other than the error callback in several places. I like the addition of being able to define edges above 1. That is definitely a welcome addition

Levi-Armstrong:
Looks good. As @mpowelson mentioned we do need the ability to attache 0 edge to another process other that the error callback if that is not possible with the current implementation.

tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
```diff
@@ -47,159 +52,68 @@ TaskflowContainer GraphTaskflow::generateTaskflow(TaskInput input, TaskflowVoidF
...
-      }
-      else
-      {
-        throw std::runtime_error("Invalid Edges for process index: " + std::to_string(src_idx));
-      }
+      tasks.at(i).precede(tasks.at(static_cast<std::size_t>(idx)));
+    }
+
+    // If no edges exist for the current node, then make sure this node precedes the done task
+    if (node.edges.empty())
```
Levi-Armstrong, line 94: Should we throw and exception here and make the developer fix?

marip8:
> Does this allow you to still define 0 edges that don't go to the error callback? Sending 0 returns to error when no explicit edges are defined is fine, but I am currently sending failures of tasks to things other than the error callback in several places. I like the addition of being able to define edges above 1. That is definitely a welcome addition

I can change the assumption to be that the error callback gets added as the connection at index 0 if no other edges are available. Is that the behavior you're looking for?

tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
marip8: I don't think so. Users don't have access to the done and error tasks, so they can't explicitly add them to the graph. The implicit assumption right now is that connections are automatically made to the done and error tasks for leaf nodes

marip8:
Any ideas what this CI error means or what caused it?

`2021-01-19T16:12:24.7771847Z /root/target_ws/install/setup.bash: line 11: COLCON_TRACE: unbound variable`

Levi-Armstrong:
> Any ideas what this CI error means or what caused it?
> 
> `2021-01-19T16:12:24.7771847Z /root/target_ws/install/setup.bash: line 11: COLCON_TRACE: unbound variable`

Yea it is related to this [commit](https://github.com/ros-industrial-consortium/tesseract_planning/pull/15/commits/c022a80a5bb2877300e9619ebf4a43b8ec4e0e75). They changed something in industrial ci which broke our focal builds.

tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
Levi-Armstrong: OK, so the developer would need to visualize the graph to insure they correctly built the graph.

Levi-Armstrong:
> I can change the assumption to be that the error and done tasks only get added for leaf nodes (i.e. nodes with no edges specified). Is that the behavior you're looking for?

I think that would be sufficient. So if the leaf node was conditional it would connect to error and done task and if it was not conditional it would connect to the done callback?

marip8:
> I think that would be sufficient. So if the leaf node was conditional it would connect to error and done task and if it was not conditional it would connect to the done callback?

Exactly

marip8:
Here is the updated graph from the unit test per the changes discussed above. Only the leaf tasks are connected to the done and error callbacks

image

Levi-Armstrong:
Looks great! I restarted clang-tidy, I think we need to limit the number of process because I think it keeps running out of resources which is why we are getting random failures.

tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
```diff
@@ -96,23 +62,36 @@ class GraphTaskflow : public TaskflowGenerator
...
+   * @param dest A list of IDs of the destination nodes of the edges.
    */
-  void addEdge(int src, SourceChannel src_channel, int dest, DestinationChannel dest_channel);
+  void addEdges(int source, std::vector<int> destinations);
 
 private:
+  /** @brief Helper struct for holding relevant information about the nodes */
+  struct Node
+  {
+    Node(TaskGenerator::UPtr process_, const bool is_conditional_);
```
Levi-Armstrong, line 87: Clang-Tidy remove const before bool. Should the parameters have an underscore at the end?

marip8: I added the underscores to differentiate the arguments from the struct members [in the initializer list](https://github.com/marip8/tesseract_planning/blob/38afc2aba11cf8214bf764fc73ef77ea8eff60c2/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp#L41)

Levi-Armstrong (approved):

Levi-Armstrong:
@marip8 Are you planning to update some of the taskflow generators to leverage this?

marip8:
Yes, I can update them in subsequent PRs

mpowelson:
@marip8 So how do you get the error/done callback indices to manually connect them? For example in your second image above I want to connect A0 to the error callback instead of A00.

marip8:
Currently the error and done tasks are created inside `generateTaskflow` and *are implicitly connected only to leaf nodes*. If you wanted to connect A0 to the error callback, you would have to make it a conditional leaf node (i.e. no defined edges).

With the current implementation, I don't think it's possible to connect arbitrary nodes to the error or done task in a different way

Levi-Armstrong:
Is there any reason we not expose this? I can see a case where you have just a process which is not condition used to clean up something when an error occurs before exiting.

Levi-Armstrong:
What about creating adding the error and done call back to the node list? The 0 index node would always be the error callback and 1 index node would the done callback allow the ability to reference them when creating edges.

mpowelson:
I assume you are going to have to add that ability back in. The freespace taskflow for example requires that unless you start adding dummy tasks.
freespace_example

Levi-Armstrong:
@marip8 What do you think?

mpowelson:
Also, I realize I didn't actually review this so I should have brought it up beforehand, but I'm not sure that changing from NodeType to a bool is necessarily better. I think I'd be inclined to keep the NodeType enum just to allow us to expand it later if we want to add other things. [HERE](https://github.com/taskflow/taskflow/blob/c35cd671a4f4f838c8f29682be3fedc27b004950/taskflow/core/task.hpp#L21) are the task types for what its worth.

marip8:
I thought about this when I made this PR, and I just tried to keep the existing functionality as much as possible. What makes this difficult is that the virtual method `generateTaskflow` takes the done and error callback functions as parameters, meaning that a user can't know about the error and done tasks during the creation of the graph. To get the functionality you want, I think we would need to:
- Remove the done and error callbacks from the `generateTaskflow` function.
- Create simple tasks generators that can run a single function (the existing done and error callbacks). This possibly already exist
- Require that users add the error and done tasks to the graph manually and connect them how they want

If the plan is to use this graph taskflow generator as the basis of the other taskflow generators, changing the `generateTaskflow` function API might not be terribly painful

mpowelson:
Can we just make addNode always return a position number and make error and done special indices?

marip8:
> Can we just make addNode always return a position number and make error and done special indices?

Sure. If I'm understanding correctly, you're asking for the graph to reserve indices 0 and 1 for error and done, respectively, and all added nodes would receive indices from 2+? And then you could connect new nodes to indices 0 and 1? I think that's a reasonable solution as long as we document the implicit behavior that node indices 0 and 1 represent the error and done tasks

Levi-Armstrong:
> Sure. If I'm understanding correctly, you're asking for the graph to reserve indices 0 and 1 for error and done, respectively, and all added nodes would receive indices from 2+? And then you could connect new nodes to indices 0 and 1? I think that's a reasonable solution as long as we document the implicit behavior that node indices 0 and 1 represent the error and done tasks.

That is what I was thinking. @mpowelson what do you think?

> I assume we would still need to connect all leaf nodes to the error and done tasks still right?

That is fine with me, the another option is to leave it to the developer to do all of the connecting. I think either way they will need to dump the graph to make sure it was configured correctly and that they did not make an error.

mpowelson:
> > Sure. If I'm understanding correctly, you're asking for the graph to reserve indices 0 and 1 for error and done, respectively, and all added nodes would receive indices from 2+? And then you could connect new nodes to indices 0 and 1? I think that's a reasonable solution as long as we document the implicit behavior that node indices 0 and 1 represent the error and done tasks.
> 
> That is what I was thinking. @mpowelson what do you think?

I don't think it matters what the indices are. We could make them negative if that would make things easier so the other indices are still the location of the node in nodes_

PR #20, opened 2021-01-20
Update due to changes in tesseract_visualization interface

Levi-Armstrong:
This update to the latest visualization interface in Tesseract. It also switches the default for descartes to use floats.

### DIFF
--- a/tesseract_motion_planners/src/core/utils.cpp
+++ b/tesseract_motion_planners/src/core/utils.cpp
@@ -40,6 +40,133 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_planning
 {
+Eigen::Isometry3d calcPose(const Waypoint& wp,
+                           const std::string& working_frame,
+                           const std::string& tip_link,
+                           const Eigen::Isometry3d& tcp,
+                           const tesseract_environment::EnvState::ConstPtr& current_state,
+                           tesseract_environment::StateSolver& state_solver)
+{
+  if (isStateWaypoint(wp))
+  {
+    const auto* swp = wp.cast_const<StateWaypoint>();
+    assert(static_cast<long>(swp->joint_names.size()) == swp->position.size());
+    tesseract_environment::EnvState::Ptr state = state_solver.getState(swp->joint_names, swp->position);
+    return (state->link_transforms[tip_link] * tcp);
+  }
+
+  if (isJointWaypoint(wp))
+  {
+    const auto* jwp = wp.cast_const<JointWaypoint>();
+    assert(static_cast<long>(jwp->joint_names.size()) == jwp->size());
+    tesseract_environment::EnvState::Ptr state = state_solver.getState(jwp->joint_names, *jwp);
+    return (state->link_transforms[tip_link] * tcp);
+  }
+
+  if (isCartesianWaypoint(wp))
+  {
+    const auto* cwp = wp.cast_const<CartesianWaypoint>();
+    if (working_frame.empty())
+      return (*cwp);
+
+    return (current_state->link_transforms.at(working_frame) * (*cwp));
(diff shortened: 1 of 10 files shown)


tesseract_motion_planners/src/core/utils.cpp
```diff
@@ -40,6 +40,150 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
+          poses.push_back(*cwp);
+        }
+        else
+        {
+          poses.push_back(state->link_transforms.at(manip_info.working_frame) * (*cwp));
+        }
+      }
+      else
+      {
+        CONSOLE_BRIDGE_logError("toToolpath: Unsupported Waypoint Type!");
```
marip8, line 124: Would we want a hard failure (exception) here instead? Or is this function primarily intended for visualization purposes?

tesseract_motion_planners/src/core/utils.cpp
```diff
@@ -40,6 +40,150 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
 {
+tesseract_common::Toolpath toToolpath(const Instruction& instruction,
+                                      const tesseract_environment::Environment::ConstPtr& env)
+{
+  using namespace tesseract_planning;
+
+  if (env == nullptr)
+    throw std::runtime_error("toToolpath: The environment is a nullptr!");
+
+  tesseract_common::Toolpath toolpath;
```
marip8, line 51: Out of curiosity, does a vector containing a vector of Eigen objects also need the special aligned allocator? I was under the impression that only the inner vector needs it

marip8 (commented):
I think this looks good overall; I just left some general comments. You might consider splitting the Descartes changes out into a separate PR for clarity since it addresses something totally different

tesseract_motion_planners/src/core/utils.cpp
```diff
@@ -40,6 +40,150 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
+      }
+      else
+      {
+        poses.push_back(state->link_transforms.at(manip_info.working_frame) * (*cwp));
+      }
+    }
+    else
+    {
+      CONSOLE_BRIDGE_logError("toToolpath: Unsupported Waypoint Type!");
+    }
```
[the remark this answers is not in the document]
Levi-Armstrong: I broke it out into a separate function.

tesseract_motion_planners/src/core/utils.cpp
Levi-Armstrong: I think an exception is good.

tesseract_motion_planners/src/core/utils.cpp
Levi-Armstrong: Yea anything that contains an Eigen aligned object must also use the aligned operator. Same reason classes that include member objects that require the Eigen alignment then you also have to have it in your class. It is infections.

Levi-Armstrong:
@marip8 CI passed. You good with the changes?

Levi-Armstrong:
I will clean up the commit history before merging.

marip8 (approved):

mpowelson:
@Levi-Armstrong Care to comment for posterity why you changed the default to float for descartes? It sounds like a reasonable thing to do, but I don't see the reasoning mentioned.

mpowelson:
@Levi-Armstrong Care to comment for posterity why you changed the default to float for descartes? It sounds like a reasonable thing to do, but I don't see the reasoning mentioned.

Levi-Armstrong:
In the case of Tesseract all of the calculations are still done using double precision and Descartes is just storing the joint solution and cost using floating point which does not seem to change the accuracy of the results with reason. I think we could event get away with using Eigen::Half.

PR #23, opened 2021-01-22
Port TOTG

### LINKED ISSUE #16: Integrate TOTG
Integrated into MoveIt [HERE](https://github.com/ros-planning/moveit/pull/1365/files). Looks like it is mostly Eigen types and should not be difficult to port.

mpowelson:
Port TOTG from MoveIt to Tesseract.

Closes #16


tesseract_process_managers/include/tesseract_process_managers/task_generators/time_optimal_trajectory_generation_task_generator.h
```diff
@@ -0,0 +1,94 @@
+/**
+ * @file time_optimal_trajectory_generation_task_generator.h
+ * @brief Perform TOTG
+ *
+ * @author Levi Armstrong
+ * @date August 11. 2020
+ * @version TODO
+ * @bug No known bugs
```
Levi-Armstrong, line 8: Add your name and new date?

tesseract_process_managers/src/task_generators/time_optimal_trajectory_generation_task_generator.cpp
```diff
@@ -0,0 +1,153 @@
+/**
+ * @file time_optimal_trajectory_generation_task_generator.h
+ * @brief Perform TOTG
+ *
+ * @author Levi Armstrong
+ * @date August 11. 2020
+ * @version TODO
+ * @bug No known bugs
```
Levi-Armstrong, line 8: Same here?

tesseract_process_managers/src/task_generators/time_optimal_trajectory_generation_task_generator.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+//    // If there is a move profile associated with it, override the parameters
+//    auto it = move_profiles.find(profile);
+//    if (it != move_profiles.end())
+//    {
+//      velocity_scaling_factors[idx] = it->second->max_velocity_scaling_factor;
+//      acceleration_scaling_factors[idx] = it->second->max_acceleration_scaling_factor;
+//    }
+//  }
+  /// @todo This algorithm does not currently support different scalings for different parts of the trajectory
```
Levi-Armstrong, line 123: Do you think this is possible implement later?

tesseract_time_parameterization/include/tesseract_time_parameterization/time_optimal_trajectory_generation.h
```diff
@@ -0,0 +1,197 @@
...
+                         double max_acceleration_scaling_factor = 1.0) const;
+
+private:
+  const double path_tolerance_;
+  const double resample_dt_;
+  const double min_angle_change_;
+};
+}  // namespace tesseract_planning
+
+namespace totg
```
Levi-Armstrong, line 72: Should we just encapsulate this under the tesseract_planning namespace and remove totg?

Levi-Armstrong (commented):
I am not requesting that we do this now but do you think it would be good to break the different classes into separate files?

tesseract_process_managers/src/task_generators/time_optimal_trajectory_generation_task_generator.cpp
mpowelson: Maybe. However, I think we may be better off implementing that ourselves as a post processing step just scaling everything by the factor.

tesseract_time_parameterization/include/tesseract_time_parameterization/time_optimal_trajectory_generation.h
mpowelson: I actually added that namespace because I thought Trajectory, Path, etc were pretty generic, and I didn't want there to be clashes. A user should never really touch those anyway.

Levi-Armstrong: Makes sense. It might be good move togo namespace inside the tesseact_planning namespace?

tesseract_process_managers/src/task_generators/time_optimal_trajectory_generation_task_generator.cpp
mpowelson: This is currently prevented by the fact that this algorithm results in a flattened composite. I will look into preserving the subcomposite structure. That is marked as another todo.

mpowelson: This is currently prevented by the fact that this algorithm results in a flattened composite. I will look into preserving the subcomposite structure. That is marked as another todo.

tesseract_process_managers/include/tesseract_process_managers/task_generators/time_optimal_trajectory_generation_task_generator.h
mpowelson: Done

tesseract_time_parameterization/include/tesseract_time_parameterization/time_optimal_trajectory_generation.h
mpowelson: Sure. I can do that.

mpowelson:
Rebased after adding command language to this repo.

Levi-Armstrong:
Hopefully now most of the development will be happening in this repository.

mpowelson:
> I am not requesting that we do this now but do you think it would be good to break the different classes into separate files?

Just saw this. Maybe. I think an ideal situation would be that we move these algorithms into their own repo similar to how planners are their own repos. Then we refactor them to use a common interface and add an interface here to the algorithms and submit a PR to MoveIt to use them as well.

mpowelson:
Feel free to squash this when we merge it

mpowelson:
Force pushed to fix a typo in the commit message.

Levi-Armstrong:
> I think an ideal situation would be that we move these algorithms into their own repo similar to how planners are their own repos. Then we refactor them to use a common interface and add an interface here to the algorithms and submit a PR to MoveIt to use them as well.

Sounds good to me.

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
```diff
@@ -0,0 +1,996 @@
...
+
+  // Compute sample count
+  size_t sample_count = static_cast<std::size_t>(std::ceil(parameterized.getDuration() / resample_dt_));
+
+  // Resample and fill in trajectory
+  auto input_instruction = trajectory[0].get().cast<MoveInstruction>();
+  CompositeInstruction new_program(program);
+  new_program.clear();
+
+  /// @todo Figure out how to maintain the original seed structure of subcomposites
```
mpowelson, line 180: @Levi-Armstrong I looked into this. I don't think this will be possible without substantial modification. There are several layers we would have to make sure we pass this information through correctly. This also means that applying scaling factors on a PlanInstruction basis won't be possible.

PR #26, opened 2021-01-24
Remove tcmalloc as a library dependency

### LINKED ISSUE #25: Don't use tcmalloc as a library dependency
`tcmalloc` should only be a dependency of executables, not libraries, because different executables have their own memory management strategies. Right now Python crashes when loading the tesseract_process_managers module because it appears to be incompatible with Python.

johnwason:
tcmalloc should only be used when linking an executable.

This fixes #25

### DIFF
--- a/tesseract_process_managers/CMakeLists.txt
+++ b/tesseract_process_managers/CMakeLists.txt
@@ -19,10 +19,6 @@ find_package(tesseract_command_language REQUIRED)
 find_package(tesseract_motion_planners REQUIRED)
 find_package(tesseract_time_parameterization REQUIRED)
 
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
So this means that anyone that plans to use the library needs to link there executable against tcmalloc to get the memory advantages? We were hoping to avoid this because there is a significant benefit to using this with the planning server library.

johnwason:
Yeah, it needs to be the final executable that decides to use the tcmalloc. Otherwise it is likely to conflict with the memory management of whatever the runtime is using.

Levi-Armstrong (approved):

PR #30, opened 2021-01-27
Add overrides to Instructions and TOTG rescaling

mpowelson:
The first commit adds profile overrides to Plan, Move, and Composite instructions. These will override whatever is in the task profile dictionary if they are present.

These second commit adds rescaling to TOTG. If a move profile is present it sets scaling to 1 and then rescales after the fact.

Also closes #30


tesseract_command_language/include/tesseract_command_language/composite_instruction.h
```diff
@@ -65,6 +65,24 @@ class CompositeInstruction
   void setProfile(const std::string& profile);
   const std::string& getProfile() const;
 
+  /** @brief Add a profile object that will override the named profile for a specific task
+   * @param task_name Name of the task to which this profile will be applied
+   * @param profile ConstPtr to profile of appropriate type for the given task  */
+  void addProfileOverride(const std::string& task_name, std::shared_ptr<const void> profile);
```
Levi-Armstrong, line 71: Should we use the profile dictionary here for storing the profiles?

mpowelson: Probably. I forgot that existed. We'll have to move it into command language. Do you want me to just add a ProfileDictionary::Ptr as a public member?

Levi-Armstrong: I believe it may be generic enough to go in tesseract_common?

> Do you want me to just add a ProfileDictionary::Ptr as a public member?

I think that should be fine. It may be worth checking the cpp core guidelines.

mpowelson:
@Levi-Armstrong I'm going to proceed with moving the profile dictionary to tesseract_common and adding one to each instruction for overrides. However, I did stumbled upon [these](https://spec.oneapi.com/versions/latest/elements/oneTBB/source/containers.html). These might be useful to us in the future. oneTBB looks to be a Taskflow competitor, but we could probably borrow the containers if we wanted to.

tesseract_command_language/include/tesseract_command_language/composite_instruction.h
mpowelson: I am making it a public ProfileDictionary::Ptr as indicated by [C131](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c131-avoid-trivial-getters-and-setters).

Also, I moved it into tesseract_command_language rather than tesseract_common since I figured that might be more likely to change and it is unlikely to be used in tesseeract core.

mpowelson:
@Levi-Armstrong This is ready for review again.

Levi-Armstrong:
I have not looked into the details but how does the override profiles work? If an instruction has a profile dictionary does it first look there for the profile then if not found it checks the main profile dictionary?

mpowelson:
> I have not looked into the details but how does the override profiles work? If an instruction has a profile dictionary does it first look there for the profile then if not found it checks the main profile dictionary?

Effectively yes. In implementation I think it actually looks in the main dictionary first. However, the ones in the instruction take precedence over the main dictionary

Levi-Armstrong (approved):

PR #31, opened 2021-01-28
This fixes several items, see commit comments

Levi-Armstrong:


tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
```diff
@@ -289,7 +289,7 @@ class JointWaypoint
 
     // Check if they are close to 0
     Eigen::VectorXd range = upper_tolerance - lower_tolerance;
-    double sum = range.sum();
+    double sum = range.cwiseAbs().sum();
     return (sum > 1e-5);
```
marip8, line 293: Why don't you use `Eigen::isApprox(upper_tolerance.cwiseAbs(), lower_tolerance.cwiseAbs())`? The `1e-5` seems arbitrary; if you care about them being the same you should use machine `epsilon` which is think is the default for `Eigen::isApprox`

README.md
```diff
@@ -49,7 +49,19 @@ The planning server is built leveraging the library taskflow which enables you t
 
 ### Ubuntu
 
-This was addressed by leveraging tcmalloc library which is part of google performance tools (gperftools). This has shown to provided more stable memory management when leveraging the planning server.
+### Recommend (tcmalloc)
```
marip8, line 52:
```suggestion
### `tcmalloc` (recommended)
```

README.md
```diff
@@ -49,7 +49,19 @@ The planning server is built leveraging the library taskflow which enables you t
...
+
+The recommended approach to address this is by leveraging tcmalloc library which is part of google performance tools (gperftools). This has shown to provided more stable memory management when leveraging the planning server.
```
marip8, line 54:
```suggestion
The recommended approach to address memory management is to leverage the `tcmalloc` library which is part of Google Performance Tools (`gperftools`). Experimentation has shown `tcmalloc` to provide more stable memory management when leveraging the planning server.
```

README.md
```diff
@@ -49,7 +49,19 @@ The planning server is built leveraging the library taskflow which enables you t
...
+
+This can be leveraged by linking your exectuable which leverages the planning server against tcmalloc by doing the following:
```
marip8, line 56:
```suggestion
This capability can be leveraged by linking the exectuable that contains the planning server against `tcmalloc` using the following commands:
```

README.md
```diff
@@ -61,6 +73,26 @@ TCMALLOC_RELEASE_RATE 	default: 1.0 	Rate at which we release unused memory to t
 
 Note: This is typically set in the roslaunch file for the planning server node leveraging the `<env name="TCMALLOC_RELEASE_RATE" value="10">` tag.
 
+### Default Ubuntu OS malloc (ptmalloc)
```
marip8, line 76:
```suggestion
### `ptmalloc` (Ubuntu OS default)
```

README.md
```diff
@@ -61,6 +73,26 @@ TCMALLOC_RELEASE_RATE 	default: 1.0 	Rate at which we release unused memory to t
...
+
+The great thing is, the default malloc library `ptmalloc` used provided run-time [tunables](https://www.gnu.org/software/libc/manual/html_node/Tunables.html), leveraging a set of environment variables.
```
marip8, line 78:
```suggestion
`ptmalloc` provides run-time [tunables](https://www.gnu.org/software/libc/manual/html_node/Tunables.html) through a set of environment variables.
```

README.md
```diff
@@ -61,6 +73,26 @@ TCMALLOC_RELEASE_RATE 	default: 1.0 	Rate at which we release unused memory to t
...
+
+The one parameter that has been used to configure libc so it always gives back the memory to the system is setting following memory tunnable:
```
marip8, line 80: You should elaborate on why giving memory back to the system is important and why it doesn't do it by default.

Also `tunnable` -> `tunable`

tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
Levi-Armstrong: > Eigen::isApprox(upper_tolerance.cwiseAbs(), lower_tolerance.cwiseAbs())

We probably don't want to use cwiseAbs() because -pi would show equal to pi, but I see where you are going.

Suggest changing to: `Eigen::isApprox(upper_tolerance, lower_tolerance)`?

README.md
Levi-Armstrong: I believe the main paragraph under `Process Planning Server Tuning` explains this, does it not?

See below:

The planning server is built leveraging the library taskflow which enables you to quickly write parallel and heterogeneous tasks programs which aligns well with process motion planning. These types of libraries leverage thread pools to quickly spin up different tasks as needed. One interesting thing found, if a task consumes a lot of memory libc will make a performance decision to allow the thread to hold onto the memory it has for subsequent tasks opposed to giving the memory back to your system. This may be fine if this was the only set of tasks being run but typically in our use case there are other computationally intensive processes which could use this memory.

marip8: Okay I think that's fine. I must have skipped over that section

Levi-Armstrong:
@marip8 I made the requested changes but have two open items left to discuss. I will clean up the commit history once approved.

tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
marip8: I was thinking that `lower_tolerance` might be negative and need to have the absolute value taken. If both vectors are positive then that would work

README.md
Levi-Armstrong: Yea it most likely was not in the changes because I don't think I changed this paragraph.

tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
```diff
@@ -287,10 +287,8 @@ class JointWaypoint
     if (lower_tolerance.size() == 0 || upper_tolerance.size() == 0)
       return false;
 
-    // Check if they are close to 0
-    Eigen::VectorXd range = upper_tolerance - lower_tolerance;
-    double sum = range.sum();
-    return (sum > 1e-5);
+    // Check if they are the same
+    return Eigen::IsApprox(lower_tolerance, upper_tolerance);
```
marip8, line 291: Is this returning the opposite of what it did before? It seems like previously true meant that the values were different (i.e.  `sum` was larger than the tolerance) and now true means that they are the same

Levi-Armstrong: Yep good catch it should be `!Eigen::IsApprox(lower_tolerance, upper_tolerance);`

tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
Levi-Armstrong: I just assume that `Eigen::isApprox` handles this when checking against the epsilon but not sure if that is true.

Levi-Armstrong: It looks like the isApprox function should handle this; do you agree?

The fuzzy compares are done multiplicatively. Two vectors v and w are considered to be approximately equal within precision p if

∥v−w∥⩽pmin(∥v∥,∥w∥).
For matrices, the comparison is done using the Hilbert-Schmidt norm (aka Frobenius norm L2 norm).
Because of the multiplicativeness of this comparison, one can't use this function to check whether *this is approximately equal to the zero matrix or vector. Indeed, isApprox(zero) returns false unless *this itself is exactly the zero matrix or vector. If you want to test whether *this is zero, use internal::isMuchSmallerThan(const RealScalar&, RealScalar) instead.

Levi-Armstrong: Now that I read the description I am not sure we should use this check by itself because of the zero case.

marip8: It appears that `isApprox` doesn't do element-wise comparison, and it also doesn't make sure both vectors are positive. [Here is the description](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ae8443357b808cd393be1b51974213f9c)

marip8: I don't think this will work if one of the vectors is negative because of the subtraction on the left hand side. If you're going to use this, the vectors both need to be positive. Also if they can be zero vectors (which I assume they are by default), then the original change might be better (with the addition of a comparison to machine epsilon instead of 1e-5)

Levi-Armstrong: This approach `Eigen::isApprox(upper_tolerance.cwiseAbs(), lower_tolerance.cwiseAbs())` fails in the unit test. The following `!Eigen::IsApprox(lower_tolerance, upper_tolerance)` works but based on the description it does pose concern for the all zero case. I created a vectorized version of `almostEqualRelativeAndAbs` which is probably the best solution.

``` c++
  std::vector<std::string> joint_names = {"joint_1", "joint_2", "joint_3"};
  Eigen::VectorXd joint_values = Eigen::VectorXd::Constant(3, 1);

  JointWaypoint jw(joint_names, joint_values);
  EXPECT_FALSE(jw.isToleranced());

  jw.upper_tolerance = Eigen::VectorXd::Constant(3, 5);
  jw.lower_tolerance = Eigen::VectorXd::Constant(3, -5);
  EXPECT_TRUE(jw.isToleranced());

  jw.upper_tolerance = Eigen::VectorXd::Constant(3, 0);
  jw.lower_tolerance = Eigen::VectorXd::Constant(3, 0);
  EXPECT_FALSE(jw.isToleranced());
```

Levi-Armstrong: I update the PR to leverage the `almostEqualRelativeAndAbs` function.

tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
Levi-Armstrong: updated to leverage `almostEqualRelativeAndAbs`

Levi-Armstrong:
This PR depend on this tesseract [PR](https://github.com/ros-industrial-consortium/tesseract/pull/536).

Levi-Armstrong:
@marip8 and @mpowelson This PR good to merge?

tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
```diff
@@ -287,10 +287,9 @@ class JointWaypoint
     if (lower_tolerance.size() == 0 || upper_tolerance.size() == 0)
       return false;
 
-    // Check if they are close to 0
-    Eigen::VectorXd range = upper_tolerance - lower_tolerance;
-    double sum = range.sum();
-    return (sum > 1e-5);
+    // Check if they are the same
+    static auto max_diff = static_cast<double>(std::numeric_limits<float>::epsilon());
+    return !tesseract_common::almostEqualRelativeAndAbs(lower_tolerance, upper_tolerance, max_diff);
```
mpowelson, line 292: The same code is present in cartesian_waypoint.

mpowelson (approved):

tesseract_command_language/test/joint_waypoint_test.cpp
```diff
@@ -0,0 +1,56 @@
...
+TEST(TesseractCommandLanguageJointWaypointUnit, isToleranced)
+{
+  std::vector<std::string> joint_names = { "joint_1", "joint_2", "joint_3" };
+  Eigen::VectorXd joint_values = Eigen::VectorXd::Constant(3, 1);
+
+  JointWaypoint jw(joint_names, joint_values);
+  EXPECT_FALSE(jw.isToleranced());
+
+  jw.upper_tolerance = Eigen::VectorXd::Constant(3, 5);
+  jw.lower_tolerance = Eigen::VectorXd::Constant(3, -5);
```
marip8, line 43: I think this technically violates your description of the tolerances:

>   /** Joint distance below waypoint that is allowed. **Each element should be <= 0** */
  Eigen::VectorXd lower_tolerance;

marip8:
The descriptions of upper and lower tolerances seem a bit ambiguous to me. I think the way that you are intending to use them is similar to mechanical drawings where tolerances can be positive or negative and they are always added to the nominal value to get the toleranced value. Here are some simple examples, each of which is valid

```
5 + [-0.1, +0.1] = [4.9, 5.1] -> is toleranced
5 + [+0.1, +0.2] = [5.1, 5.1] -> is toleranced
5 + [-0.2, -0.1] = [4.8, 4.9] -> is toleranced
5 + [+0.1, +0.1] = [5.1, 5.1] -> is **not** toleranced
```

I think the checks you need to make are:
1. `lower_tolerance` is always element-wise smaller than `upper_tolerance`
2. If `lower_tolerance` is equal to `upper_tolerance

I would also suggest updating the documentation to make these assumptions about the tolerances clear

tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
marip8, line 292:
```suggestion
    // Check if the upper tolerance is greater than the lower tolerance by some small amount
    return ((upper_tolerance - lower_tolerance).array() > max_diff).all();
```

tesseract_command_language/test/joint_waypoint_test.cpp
Levi-Armstrong: I am not following how this violate the description; is there something I am missing? The lower tolerance is being populated with a vector of length 3 with a constant value of `-5`. All values in the vector should be `-5` which is `<=0`.

marip8: Sorry, I misread this. Disregard my previous comment

tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
Levi-Armstrong: added

Levi-Armstrong: @marip8 The changes discussed in our meeting have been applied. Let me know if you see an issues with the changes.

marip8 (approved):

marip8:
Just to summarize our earlier discussion:
- Lower tolerance is a set of numbers that should be element-wise <= 0.0
- Upper tolerance is a set of numbers that should be element-wise >= 0.0
- The joint state limit is calculated by adding the tolerance to the nominal value

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
```diff
@@ -277,13 +277,17 @@ class CircularPathSegment : public PathSegment
...
+    // This does not match the paper which should be acos(start_direction.dot(end_direction))
     const double angle = acos(std::max(-1.0, start_direction.dot(end_direction)));
     const double start_distance = (start - intersection).norm();
     const double end_distance = (end - intersection).norm();
 
     // enforce max deviation
-    double distance = std::min(start_distance, end_distance);
-    distance = std::min(distance, max_deviation * sin(0.5 * angle) / (1.0 - cos(0.5 * angle)));
+    double l1 = start_distance * 0.5;
+    double l2 = end_distance * 0.5;
```
Levi-Armstrong, line 287: The paper was multiplying by 0.5.

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
```diff
@@ -845,7 +863,7 @@ double Trajectory::getMinMaxPathAcceleration(double path_position, double path_v
   double max_path_acceleration = std::numeric_limits<double>::max();
   for (unsigned int i = 0; i < joint_num_; ++i)
   {
-    if (config_deriv[i] != 0.0)
+    if (!tesseract_common::almostEqualRelativeAndAbs(config_deriv[i], 0.0, std::numeric_limits<double>::epsilon()))
```
Levi-Armstrong, line 866: This change seems to make the trajectory duration increase for the tests. It was dividing by very small numbers because they never actually equaled zero.

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
Levi-Armstrong: Also this change makes the trajectory duration increase. Not sure why it is different. 

From the [paper](http://www.roboticsproceedings.org/rss08/p27.pdf):

>where  the  first  two  elements  give  the  maximum  possible distances such that the circular segment does not replace more than half of the adjoining linear segments and the last element limits  the  radius  to  make  sure  the  circlular  segment  stays within a distance δ from waypoint qi

Levi-Armstrong: @mpowelson let me know if you have an issue with this change and I can change it back.

mpowelson: It's possible that these changes were added by the move it folks. I'm not sure how closely this follows the original implementation.

gavanderhoorn: The PR which added it to MoveIt was pretty well structured. See https://github.com/ros-planning/moveit/pull/1365.

The first commit imports the upstream code, directly from the author of the paper. Later commits add the wrappers and MoveIt adapters.

Levi-Armstrong: @mpowelson The original commit form the author had it without the 0.5 factor so I have reverted my change to keep it the same.

@gavanderhoorn Thank you for the information and I was able to found the commit which added the change `acos(std::max(-1,...` so I added the comments from the PR in the code for future readers if they compare back to the paper.

gavanderhoorn: I would perhaps even just straight link to the MoveIt PRs.

And if we fix something here, contribute it back.

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
```diff
@@ -818,10 +827,24 @@ void Trajectory::integrateBackward(std::list<TrajectoryStep>& start_trajectory,
     // Check for intersection between current start trajectory and backward
     // trajectory segments
     const double start_slope = (start2->path_vel_ - start1->path_vel_) / (start2->path_pos_ - start1->path_pos_);
-    const double intersection_path_pos =
-        (start1->path_vel_ - path_vel + slope * path_pos - start_slope * start1->path_pos_) / (slope - start_slope);
-    if (std::max(start1->path_pos_, path_pos) - EPS <= intersection_path_pos &&
-        intersection_path_pos <= EPS + std::min(start2->path_pos_, trajectory.front().path_pos_))
+
+    // It is possible to have both slop and start_slope to be equal
```
gavanderhoorn, line 831:
```suggestion
    // It is possible to have both slope and start_slope to be equal
```

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
```diff
@@ -818,10 +827,24 @@ void Trajectory::integrateBackward(std::list<TrajectoryStep>& start_trajectory,
...
+    bool check_eq_slope = tesseract_common::almostEqualRelativeAndAbs(slope, start_slope, EPS);
+    double intersection_path_pos{ 0 };
+    if (check_eq_slope)
+      intersection_path_pos = start1->path_pos_ + (start2->path_pos_ - start1->path_pos_) / 2.0;
+    else
+      intersection_path_pos =
+          (start1->path_vel_ - path_vel + slope * path_pos - start_slope * start1->path_pos_) / (slope - start_slope);
```
gavanderhoorn, line 838: this could perhaps do with a comment? What happens if they are equal? And why? What if they are not?

Levi-Armstrong: > What happens if they are equal? 

This occurs if two consecutive `TrajectorySteps` have the same acceleration. Resulting in `intersection_path_pos` being nan because it divides by zero.

> What if they are not?

I have not looked at the paper to see what is being calculated here.

gavanderhoorn: I (perhaps wrongfully) assumed this was added, due to the `+` in the gutter.

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
Levi-Armstrong: > I would perhaps even just straight link to the MoveIt PRs.

Done.

> And if we fix something here, contribute it back.

Ideally that would be preferred, though I personally don't have the bandwidth for this. Though my preference would be to create a separate repository for time parameterization and join resource with maintaining it since these are independent of Tesseract and MoveIt. Not sure if the maintainers of MoveIt would be open to this, but I would be.

Issue #27, opened 2021-01-25
TOTG Error: Error while integrating backward: Negative path velocity


mpowelson:
I assume this is related to https://github.com/ros-planning/moveit/issues/2495. Wherever this gets addressed first, we need to port the fix or post it over there.

Levi-Armstrong:
This issues is most likely due to uninitialized data [here](https://github.com/ros-industrial-consortium/tesseract_planning/blob/2c61eba797a158eb5775f44adaf15e0d6426c9a2/tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp#L774) which can used. The question is what should it be initialized to?

Levi-Armstrong:
Also noticed that there is several `<=` and `>=` when comparing floats. Some places it leverages and EPS to account for numerical issues when comparing floats, but not all places.

mpowelson:
Odd the clang tidy build didn't fail. Yeah. I also seem to remember seeing somewhere where I thought there was a memory leak. This implementation probably needs a general cleanup

mpowelson:
There is an assert checking the condition where it would be unitialized. I'm actually getting this pretty frequently now. However, that assert is not triggered when it happens

mpowelson:
I'm not sure of the root cause, but it seems that a workaround is to append an extra joint that just increments such that no point ever repeats.

This passes
```
TEST(time_optimal_trajectory_generation, test_return_home)
{
  Eigen::VectorXd waypoint(7);
  std::list<Eigen::VectorXd> waypoints;

  waypoint << 0, 0.7, -2.1, 0, -0.25, 0,1;
  waypoints.push_back(waypoint);
  waypoint << 0, 0, 0, 0, 0, 0,2;
  waypoints.push_back(waypoint);
  waypoint << 0, 0.70001, -2.1, 0, -0.25, 0,3;
  waypoints.push_back(waypoint);
  waypoint << 0, 0, 0, 0, 0, 0.1, 4;
  waypoints.push_back(waypoint);

  Eigen::VectorXd max_velocities(7);
  max_velocities.setOnes();
  Eigen::VectorXd max_accelerations(7);
  max_accelerations.setOnes();

  Trajectory trajectory(Path(waypoints, 0.001), max_velocities, max_accelerations, 0.001);
  EXPECT_TRUE(trajectory.isValid());
}
```

mpowelson:
[...]

PR #32, opened 2021-01-28
Add workaround for TOTG failure on duplicate points

mpowelson:
Workaround for #27 .

### DIFF
--- a/tesseract_time_parameterization/test/time_optimal_trajectory_generation_tests.cpp
+++ b/tesseract_time_parameterization/test/time_optimal_trajectory_generation_tests.cpp
@@ -38,6 +38,7 @@
 
 #include <gtest/gtest.h>
 #include <tesseract_time_parameterization/time_optimal_trajectory_generation.h>
+#include <tesseract_command_language/command_language.h>
 
 using tesseract_planning::totg::Path;
 using tesseract_planning::totg::Trajectory;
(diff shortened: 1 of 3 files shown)


Levi-Armstrong:
Does this not affect the the time parallelization results? 
If is does, I recently added a vectorized version of the almost relative and absolute equal function. This could be used to remove duplicate states and and then insert them back in.

mpowelson:
> Does this not affect the the time parallelization results?
> If it does, I recently added a vectorized version of the almost relative and absolute equal function. This could be used to remove duplicate states and and then insert them back in.

I don't think it should, but I could be wrong. Since the limits are so high it should never be the limiting joint. I don't think you can remove the point. It's not just duplicate sequential points, it is points later in the trajectory that are actually valid.

Levi-Armstrong:
Ah, so this is because it visits the same location twice in the same trajectory?

tesseract_time_parameterization/test/time_optimal_trajectory_generation_tests.cpp
```diff
@@ -228,6 +229,78 @@ TEST(time_optimal_trajectory_generation, testLargeAccel)
...
+//  waypoints.push_back(waypoint);
+
+//  Eigen::VectorXd max_velocities(7);
+//  max_velocities.setOnes();
+//  Eigen::VectorXd max_accelerations(7);
+//  max_accelerations.setOnes();
+
+//  Trajectory trajectory(Path(waypoints, 0.001), max_velocities, max_accelerations, 0.001);
+//  EXPECT_TRUE(trajectory.isValid());
+//}
```
Levi-Armstrong, line 255: It looks likes this was used to test the approach you applied; should it be removed?

mpowelson: This test still fails because of the bug in trajectory. The workaround only is applied when using the computeTimesteps function. I can remove it if you want. I left it in there so we could test if any of the changes you were making fixed the underlying issue.

mpowelson: > Ah, so this is because it visits the same location twice in the same trajectory?

I think it is just if it revisits the start or end location.

Levi-Armstrong: > This test still fails because of the bug in trajectory.

Is this bug captured somewhere; I just do not want to loose track of it? It may also be good to leverage the doxygen "@bug" functionality so it also shows up in the documentation. 

>  I left it in there so we could test if any of the changes you were making fixed the underlying issue.

I am good with it staying.

Levi-Armstrong:
I think the only thing left is to add a doxygen bug description, then this can be merged.

mpowelson:
> I think the only thing left is to add a doxygen bug description, then this can be merged.

I added it to the doxygen for Trajectory

Levi-Armstrong (approved):

PR #36, opened 2021-02-11
Add Tolerances to joint/cartesian waypoint XML

mpowelson:
Tolerances were added to the JointWaypoint and CartesianWaypoint a while back, but they were not added to the XML serialization. This adds serialization for them.


Levi-Armstrong:
Looks good. Is there a unit test that confirms the serialization is working?

mpowelson:
I changed one of the waypoints in the test to a joint waypoint so that it gets tested too. However, the serialization tests don't really test much. They just test that a composite remains the same size after serializing and deserializing. They don't actually test for correctness. I have run the serialization locally and it seems correct, but I did not try deserialization outside of the unit test.

We definitely need to improve this area of our test coverage. What would you think about adding an equality operator to the command language classes and using that in the tests?

Levi-Armstrong:
> We definitely need to improve this area of our test coverage. What would you think about adding an equality operator to the command language classes and using that in the tests?

Sounds good to me.

mpowelson:
@Levi-Armstrong This is ready for review.

mpowelson:
@Levi-Armstrong Does the docker have a different version of Eigen from the system version? The unit tests all pass on my local Focal machine.

Levi-Armstrong:
Example YAML to create a similar structure for xml.

```
namespace YAML
{

template <>
struct convert<Eigen::Isometry3d>
{
  static Node encode(const Eigen::Isometry3d& rhs)
  {
    Node xyz;
    xyz["x"] = rhs.translation().x();
    xyz["y"] = rhs.translation().y();
    xyz["z"] = rhs.translation().z();

    const Eigen::Quaterniond q(rhs.linear());
    Node quat;
    quat["x"] = q.x();
    quat["y"] = q.y();
    quat["z"] = q.z();
    quat["w"] = q.w();

    Node node;
    node["position"] = xyz;
    node["orientation"] = quat;

    return node;
  }

  static bool decode(const Node& node, Eigen::Isometry3d& rhs)
  {
    Eigen::Vector3d xyz;
    xyz.x() = node["position"]["x"].as<double>();
    xyz.y() = node["position"]["y"].as<double>();
    xyz.z() = node["position"]["z"].as<double>();

    Eigen::Quaterniond quat;
    quat.x() = node["orientation"]["x"].as<double>();
    quat.y() = node["orientation"]["y"].as<double>();
    quat.z() = node["orientation"]["z"].as<double>();
    quat.w() = node["orientation"]["w"].as<double>();

    Eigen::Isometry3d out = Eigen::Isometry3d::Identity();
    out.translation() = xyz;
    out.linear() = quat.toRotationMatrix();
    rhs = out;
    return true;
  }
};
```

mpowelson:
> @Levi-Armstrong Does the docker have a different version of Eigen from the system version? The unit tests all pass on my local Focal machine.

This requires [Tesseract PR 549](https://github.com/ros-industrial-consortium/tesseract/pull/549)

Levi-Armstrong (approved):

mpowelson:
@Levi-Armstrong Do you want me to merge this?

Levi-Armstrong:
Sure.

PR #42, opened 2021-02-26
Allow graph taskflows that connect non-leaf nodes to error/done

### LINKED ISSUE #28: GraphTaskflow doesn't support non-leaf nodes connecting to the error task
Implement changes discussed here https://github.com/ros-industrial-consortium/tesseract_planning/pull/17

mpowelson:
Closes #28 

See the before/after of the unit test
Screenshot from 2021-02-26 10-23-36

### DIFF
--- a/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
@@ -48,6 +48,11 @@ class GraphTaskflow : public TaskflowGenerator
 public:
   using UPtr = std::unique_ptr<GraphTaskflow>;
 
+  /** @brief ID for the Done node */
+  const static int DONE_NODE = -1;
+  /** @brief ID for the Error node */
+  const static int ERROR_NODE = -2;
+
   GraphTaskflow(std::string name = "GraphTaskflow");
   ~GraphTaskflow() override = default;
   GraphTaskflow(const GraphTaskflow&) = delete;
--- a/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
+++ b/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
@@ -86,7 +86,7 @@ TaskflowContainer GraphTaskflow::generateTaskflow(TaskInput input, TaskflowVoidF
       else if (idx == ERROR_NODE)
         tasks.at(i).precede(error_task);
       else
-        CONSOLE_BRIDGE_logError("GraphTaskflow invalid edge: %d", idx);
+        throw std::runtime_error("Invalid GraphTaskflow: Node specified with invalid edge");
     }
 
     // If no edges exist for the current node, make sure it precedes the done task (and error task if conditional)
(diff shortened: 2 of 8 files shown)


tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
```diff
@@ -48,6 +48,11 @@ class GraphTaskflow : public TaskflowGenerator
 public:
   using UPtr = std::unique_ptr<GraphTaskflow>;
 
+  /** @brief ID for the Done node */
+  const static int DONE_NODE = -1;
+  /** @brief ID for the Error node */
+  const static int ERROR_NODE = -2;
```
marip8, line 54:
```suggestion
  /** @brief ID for the Done node */
  const static int DONE_NODE = 0;
  /** @brief ID for the Error node */
  const static int ERROR_NODE = 1;
```
It's probably easier for these nodes to be valid vector indices so we can index into `nodes_` without doing a bunch of special checks about whether the node index is less than zero and whether it equals one of these two numbers

tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
```diff
@@ -79,7 +79,14 @@ TaskflowContainer GraphTaskflow::generateTaskflow(TaskInput input, TaskflowVoidF
...
     {
-      tasks.at(i).precede(tasks.at(static_cast<std::size_t>(idx)));
+      if (idx >= 0)
+        tasks.at(i).precede(tasks.at(static_cast<std::size_t>(idx)));
+      else if (idx == DONE_NODE)
+        tasks.at(i).precede(done_task);
+      else if (idx == ERROR_NODE)
+        tasks.at(i).precede(error_task);
+      else
+        CONSOLE_BRIDGE_logError("GraphTaskflow invalid edge: %d", idx);
```
marip8, line 89: This change would be unnecessary if `ERROR_NODE` and `DONE_NODE` were valid vector indices rather than negative numbers

marip8 (changes requested):

tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
mpowelson: The error and done tasks are not currently added to nodes, so this allows you to index into nodes_ without subtracting 2 every time. The only check you have to do is if it is less than 0. Those cases are already handled in an ad hoc fashion, so I don't see a couple of if statements as much worse than that.

marip8: I forgot the error and done tasks are added after all of the other nodes have been added. Keeping the negative indices probably make sense then. The alternative would be adding two placeholder nodes to the `nodes_` vector, then replacing them when the actual error and done tasks get generated. That seems relatively hacky though

tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
marip8, line 89: We should throw an exception here instead of only logging the error. If the graph can't be constructed per the request, we should build an incorrect or incomplete graph

marip8: Resolving, since we're keeping the negative indices for the error and done node

mpowelson: done

marip8 (approved):

PR #44, opened 2021-03-04
Add trajectory container class to abstract command lanaguage from time parameterization

Levi-Armstrong:
The time parameterization was specific to taking a command language as input. We currently have a need to just take the results out of the sqp solver and run time parameterization. This introduces a TrajectoryContainer class which you can implement for different data structures which both of the current time parameterization implementations now support. 

Also updated time optimal implementation so it does not have to modify the results structure. @mpowelson In this case I left the computeTimeStamps which takes the CompositeInstruction until you have time to test the new implementation to avoid causing any issues if it does not perform as desired.

### DIFF
--- a/tesseract_time_parameterization/src/instructions_trajectory.cpp
+++ b/tesseract_time_parameterization/src/instructions_trajectory.cpp
@@ -27,6 +27,7 @@
 #include <tesseract_time_parameterization/instructions_trajectory.h>
 #include <tesseract_command_language/utils/utils.h>
 #include <tesseract_command_language/state_waypoint.h>
+#include <tesseract_command_language/waypoint_type.h>
 
 namespace tesseract_planning
 {
@@ -34,7 +35,7 @@ static flattenFilterFn programFlattenMoveInstructionFilter =
     [](const Instruction& i, const CompositeInstruction& /*composite*/, bool parent_is_first_composite) {
       if (isMoveInstruction(i))
       {
-        if (i.cast_const<MoveInstruction>()->isStart())
+        if (i.as<MoveInstruction>().isStart())
           return (parent_is_first_composite);
 
         return true;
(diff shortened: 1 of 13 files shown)


mpowelson:
> The time parameterization was specific to taking a command language as input. We currently have a need to just take the results out of the sqp solver and run time parameterization. This introduces a TrajectoryContainer class which you can implement for different data structures which both of the current time parameterization implementations now support.

I'll try to take a look at this some time before the weekend. It looks like you've made some pretty significant changes (/improvements). Do you think it is time to move time parameterization into its own repo and drop the "Tesseract"? This is diverging from MoveIt pretty quickly.


> Also updated time optimal implementation so it does not have to modify the results structure. @mpowelson In this case I left the computeTimeStamps which takes the CompositeInstruction until you have time to test the new implementation to avoid causing any issues if it does not perform as desired.

This is very welcome. I'll be interested to see how you did this. I have a hack in #30 that sorts the resulting points back into their original form. It may not be necessary anymore.

Levi-Armstrong:
> I'll try to take a look at this some time before the weekend. It looks like you've made some pretty significant changes (/improvements). Do you think it is time to move time parameterization into its own repo and drop the "Tesseract"? This is diverging from MoveIt pretty quickly.

I think it would be a good idea to move it to it own repository in the near future. I thing we should be able to work out an interface for the time parameterization and once we have mostly wrapped up we can break it out and do a release.

> This is very welcome. I'll be interested to see how you did this. I have a hack in #30 that sorts the resulting points back into their original form. It may not be necessary anymore.

I have to take a look at what you did also. I implemented two approaches to one which should work most of the time and if it fails it will fall into a more brute force approach.

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
```diff
@@ -215,6 +215,122 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(CompositeInstruction& pr
...
+    if (diverse_point)
+      points.push_back(position);
+  }
+
+  // Return trajectory with only the first waypoint if there are not multiple diverse points
+  if (points.size() == 1)
+  {
+    CONSOLE_BRIDGE_logDebug("Trajectory is parameterized with 0.0 dynamics since it only contains a single distinct "
+                            "waypoint.");
+    trajectory.setData(0, Eigen::VectorXd::Zero(num_joints), Eigen::VectorXd::Zero(num_joints), 0);
```
mpowelson, line 299: So this only adds time param for the first point. The old version I think only returned a single point. Perhaps we should just set the time to all 0?

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
```diff
@@ -957,6 +1073,92 @@ bool Trajectory::isValid() const { return valid_; }
...
+{
+  std::list<Trajectory::TrajectoryStep>::const_iterator it = trajectory_.begin();
+  long cnt = 0;
+  for (long i = 0; i < trajectory.size(); ++i)
+  {
+    const Eigen::VectorXd& p = trajectory.getPosition(i);
+    for (; it != trajectory_.end(); ++it)
+    {
+      Eigen::VectorXd compare_p = getPosition(it->time_).head(trajectory.dof());
+      if (!((p - compare_p).norm() > path_tolerance))
```
mpowelson, line 1086: Yeah this is what I did originally as well. However, when multiple points were within the tolerance, it assigns most of them to the next composite. That's why I started checking the error and waiting until the error stopped going down before incrementing.

mpowelson (commented):
Have you by chance compared the results between the 2 or run this on a robot?

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
mpowelson: See here if you are interested https://github.com/ros-industrial-consortium/tesseract_planning/pull/30/commits/8407043606bf6a789b7bf88b4e5e6c724cb39f67#diff-35b696ce9c09c51a59cc90167b5987ccb88660c8e4b03054a2f2d594e9a7e137R194

mpowelson:
Also, one thing I have been continually tracking down is when meta information is lost in the MoveInstructions. I for instance call ISP if TOTG fails, so we need to make sure that the MoveInstructions pass all meta information through.

tesseract_time_parameterization/src/instructions_trajectory.cpp
```diff
@@ -0,0 +1,123 @@
...
+        return true;
+      }
+
+      return false;
+    };
+
+InstructionsTrajectory::InstructionsTrajectory(std::vector<std::reference_wrapper<Instruction>> trajectory)
+  : trajectory_(std::move(trajectory))
+{
+  dof_ = trajectory_.front()
```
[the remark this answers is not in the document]
Levi-Armstrong: It now checks if empty and throws an exception. This was not the case at the time of the review, but the waypoint and instruction type cast methods now throw a bad cast exception if you try to cast it to the wrong type.

tesseract_time_parameterization/src/instructions_trajectory.cpp
```diff
@@ -0,0 +1,123 @@
...
+             .cast_const<MoveInstruction>()
+             ->getWaypoint()
+             .cast_const<StateWaypoint>()
+             ->position.rows();
+}
+
+InstructionsTrajectory::InstructionsTrajectory(CompositeInstruction& program)
+{
+  trajectory_ = flatten(program, programFlattenMoveInstructionFilter);
+  dof_ = trajectory_.front()
```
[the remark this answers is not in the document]
Levi-Armstrong: Fixed same as above.

tesseract_time_parameterization/src/iterative_spline_parameterization.cpp
```diff
@@ -303,120 +200,75 @@ bool IterativeSplineParameterization::compute(
     }
   }
 
-  if (add_points_)
```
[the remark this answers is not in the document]
Levi-Armstrong: It was to accommodate the new data structure using Eigen versus std::vector. It looks like it produces the same SingleJointTrajectory as before or do you see a difference?

tesseract_time_parameterization/src/iterative_spline_parameterization.cpp
```diff
@@ -536,34 +449,33 @@ bool IterativeSplineParameterization::compute(
...
+    // Calculate time from start
+    if (i > 0)
+      time = time + time_diff[i - 1];
+
+    // Do not process added points
+    if (add_points && (i == 1 || i == num_points - 2))
     {
-      for (unsigned int j = 0; j < num_joints; j++)
-        swp->velocity[j] = t2[j].velocities_[i];
+      time = time + time_diff[i - 1];
```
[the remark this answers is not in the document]
Levi-Armstrong: I believe this was a bug.

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
Levi-Armstrong: Yea that would be good. I have made the fix.

tesseract_time_parameterization/src/time_optimal_trajectory_generation.cpp
Levi-Armstrong: I have added in your logic to improve finding the correct point on the trajectory.

Levi-Armstrong:
@mpowelson I finally got around to making the remaining changes.

PR #45, opened 2021-03-11
Update descartes and ompl to leverage CollisionCheckConfig

Levi-Armstrong:


Levi-Armstrong:
This has been tested and verified on system.

PR #49, opened 2021-03-19
Enable OMPL to plan for paths with more than one instruction

mpowelson:
This allows OMPL to process more than one plan instruction. Currently it just serially processes them on after another. They naively pass goal of the prior segment as the start of the next. This should work for joint via points, but we probably want to do something more sophisticated in the future for cartesian via points to guarantee that the same solution is used for adjacent segments.

Attached is the freespace manager example showing OMPL planning from a joint waypoint to a cartesian waypoint and back.

https://user-images.githubusercontent.com/12128406/112055658-bfe91200-8b1c-11eb-8c8d-4c3a0d202afa.mp4


Levi-Armstrong:
@mpowelson I will review and test this today.

Levi-Armstrong (approved):

PR #57, opened 2021-03-25
Reduce limits slightly in FixStateBoundsTaskGenerator

mpowelson:
When planning with OMPL I ran into issues where a goal would be 1e-15 outside of the joint limits and it would fail. This should fix that (and for any other similarly sensitive Tasks).

### DIFF
--- a/tesseract_process_managers/include/tesseract_process_managers/task_generators/fix_state_bounds_task_generator.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/task_generators/fix_state_bounds_task_generator.h
@@ -51,6 +51,12 @@ struct FixStateBoundsProfile
 
   /** @brief Maximum amount the process is allowed to correct. If deviation is further than this, it will fail */
   double max_deviation_global = std::numeric_limits<double>::max();
+
+  /** @brief Amount to reduce the upper bounds before clamping limits. Should be > 1 */
+  double upper_bounds_reduction{ std::numeric_limits<float>::epsilon() };
+
+  /** @brief Amount to increase the lower bounds before clamping limits. Should be > 1 */
+  double lower_bounds_reduction{ std::numeric_limits<float>::epsilon() };
 };
 using FixStateBoundsProfileMap = std::unordered_map<std::string, FixStateBoundsProfile::ConstPtr>;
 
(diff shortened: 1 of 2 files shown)


tesseract_process_managers/include/tesseract_process_managers/task_generators/fix_state_bounds_task_generator.h
```diff
@@ -51,6 +51,12 @@ struct FixStateBoundsProfile
 
   /** @brief Maximum amount the process is allowed to correct. If deviation is further than this, it will fail */
   double max_deviation_global = std::numeric_limits<double>::max();
+
+  /** @brief Amount to reduce the upper bounds before clamping limits. Should be > 1 */
+  double upper_bounds_reduction{ 1e-6 };
```
Levi-Armstrong, line 56: Should we use float epsilon here as the default?

tesseract_process_managers/src/task_generators/fix_state_bounds_task_generator.cpp
```diff
@@ -85,6 +85,8 @@ int FixStateBoundsTaskGenerator::conditionalProcess(TaskInput input, std::size_t
   }
 
   auto limits = fwd_kin->getLimits().joint_limits;
+  limits.col(0) = limits.col(0).array() + cur_composite_profile->lower_bounds_reduction;
```
Levi-Armstrong, line 88: Should it only adjust those that are outside the bounds?

Levi-Armstrong (commented):
I recently ran into this same issues when passing trajectories planned with descartes using floats which has a larger epsilon than double and ompl was failing due to start state out of bounds. I plan to create utility functions enforceBounds and satisfiesBounds utility functions and it probably would be good to add a task which process the whole trajectory.

tesseract_process_managers/src/task_generators/fix_state_bounds_task_generator.cpp
mpowelson: This is reducing the bounds by a small amount. The checking whether or not it is outside of the bounds happens below

tesseract_process_managers/include/tesseract_process_managers/task_generators/fix_state_bounds_task_generator.h
mpowelson: Yeah that would probably work

tesseract_process_managers/src/task_generators/fix_state_bounds_task_generator.cpp
mpowelson: I rebased and changed this

tesseract_process_managers/include/tesseract_process_managers/task_generators/fix_state_bounds_task_generator.h
mpowelson: I rebased and changed this

PR #58, opened 2021-03-26
Enforce limits on result trajectory for all motion planners.

Levi-Armstrong:
For some planners you are able to set the float type like Descartes. This can cause numerical issues with other planners because floats have a larger epsilon that doubles. For example if you generate a trajectory with Descartes using floats it is very possible to have ompl invalid start states because its function which checks bounds uses the epsilon for double. This PR adds a step for each planner which loops over the resulting trajectory and enforce limits to prevent this issue when chaining planners together. 

~~@mpowelson Hopefully this will eliminate the need to have tasks in the taskflow to fix bounds but this does not fix the other numerical issue with collisions.~~ See comment below.


mpowelson (commented):
Looks reasonable. I think the FixStateBounds Task still serves a purpose though. It is for fixing if someone passes inputs that are outside of the limits. I know we for instance have legacy code that relies on the fact that MoveIt would clamp things outside of the limits to the limits

PR #64, opened 2021-04-02
Switch over command language to use boost serialization

Levi-Armstrong:


Levi-Armstrong:
@mpowelson and @marip8 This is ready to review if either of you have time to take a look. It strips out the custom serialization code and replaces it will boost serialization. Also thanks to @marip8 was able to clean up the type erasure for waypoint and instruction streamlining things. The only think left that I will do in separate commit switch the casting to be like the Any Type Erasure.

tesseract_command_language/include/tesseract_command_language/core/instruction.h
```diff
@@ -259,25 +282,51 @@ class Instruction
...
+    if (rhs.instruction_ == nullptr)
+      return false;
+
+    return instruction_->operator==(*rhs.instruction_);
+  }
+
+  bool operator!=(const Instruction& rhs) const { return !operator==(rhs); }
 
   template <typename T>
   T* cast()
```
marip8, line 316: Not sure if you wanted to address it in this PR, but we had discussed renaming the casting function to something like `as` or at least naming the two different versions the same thing. I think you already did this in a PR for the `any` type in `tesseract_common`, so I thought you might want to do the same here and for `Waypoint`

tesseract_command_language/include/tesseract_command_language/core/instruction.h
```diff
@@ -226,15 +243,21 @@ class Instruction
   }
 
   Instruction()  // NOLINT
-    : instruction_(std::make_unique<detail_instruction::InstructionInner<uncvref_t<NullInstruction>>>())
+    : instruction_(nullptr)
   {
   }
```
marip8, line 248: It might be worth thinking about whether a default constructible `Instruction` type eraser is something that is conceptually possible to create/use. Since you have a public default constructor, you're "forced" to make a bunch of checks internally to check that the underlying object is not a `nullptr` before performing various operations on it (casting, comparison, copying, moving, etc.). That seems like a lot of hassle and is somewhat prone to mistakes. You might considering keeping your `NullInstruction` class as the default placeholder implementation of this type eraser that you use in the rest of your code. Then you can get rid of all of those `nullptr` checks

tesseract_command_language/include/tesseract_command_language/core/waypoint.h
```diff
@@ -191,15 +208,21 @@ class Waypoint
   }
 
   Waypoint()  // NOLINT
-    : waypoint_(std::make_unique<detail_waypoint::WaypointInner<uncvref_t<NullWaypoint>>>())
+    : waypoint_(nullptr)
   {
   }
```
marip8, line 213: Same comment about moving this constructor to `private`

tesseract_command_language/include/tesseract_command_language/core/instruction.h
Levi-Armstrong: Done.

tesseract_command_language/include/tesseract_command_language/core/waypoint.h
Levi-Armstrong: I was not able to move these to private because the utility function toArchiveStringXML, fromArchiveStringXML, etc have to create a default constructable object.

tesseract_command_language/include/tesseract_command_language/core/instruction.h
Levi-Armstrong: It has been switched over to the new method `as` which returns a reference now.

tesseract_command_language/include/tesseract_command_language/core/waypoint.h
marip8: I see. My work-around was to serialize/de-serialize the type into a `std::vector`. Declaring a `std::vector` of a specific type does not require the use the default constructor and `boost` has built-in serialization for most STL containers. Depending on what you think of the value of privatizing the default constructor, a work-around like this might be worth the little bit of indirection

Levi-Armstrong:
@mpowelson and @marip8 I believe everything is now ready once CI passes.

PR #66, opened 2021-04-08
Fix passing of meta information in simple planner and min_seed_length

mpowelson:
I thought I already fixed this, but on master meta information gets lost in the simple planner and the min seed length task. This fixes it.


Levi-Armstrong (approved):
If CI passes it is good to merge.

PR #68, opened 2021-04-08
Update to taskflow 3.0

mpowelson:
It seems that not much changed that we actually use. We'll need to update tesseract_ext as well.


mpowelson:
@Levi-Armstrong How/when do you want to handle this upgrade?

Levi-Armstrong:
I believe if you remove taskflow from the CI skip keys this should build.

mpowelson:
@Levi-Armstrong Will we need to remove it from the docker?

Levi-Armstrong:
@mpowelson I pushed new dockers with taskflow 3.0.0

mpowelson:
Included in #71

PR #71, opened 2021-04-10
Feature/taskflow 3

Levi-Armstrong:


Levi-Armstrong:
@mpowelson It looks like for melodic builds it is just hanging when building tesseract_motion_planners which usually means it is running out of resources. I have reduced it to using a single thread but still not getting any further.

Levi-Armstrong:
It looks like the issue was using the latest version of cmake on melodic. I created a taskflow branch changing the cmake version back to 3.10 and everything built and now CI looks like it will finish.

PR #76, opened 2021-04-27
Set taskflows default post collision check to leverage discrete lvs

Levi-Armstrong:
The continuous collision post check is problematic when you have two links that can collide which are in close proximity to each where the beginning of a casted objects collides with the end of another casted collision object.

It also removed unused parameter in Descartes default plan profile.


Levi-Armstrong:
@mpowelson Do you have any objections with changing the post collision check type?

mpowelson (approved):

PR #82, opened 2021-05-12
Update rosinstall with tag/hash and create an unstable CI build using master branches

Levi-Armstrong:
See title.

### DIFF
--- a/.github/workflows/unstable_build.yml
+++ b/.github/workflows/unstable_build.yml
@@ -1,0 +1,59 @@
+name: Unstable-Build
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  schedule:
+    - cron: '0 5 * * *'
+
+jobs:
+  industrial_ci:
+    name: Unstable
+    runs-on: ubuntu-latest
+    env:
+      CI_NAME: Unstable-Build
+      OS_NAME: ubuntu
+      OS_CODE_NAME: focal
+      ROS_DISTRO: noetic
+      ROS_REPO: main
+      UPSTREAM_WORKSPACE: 'dependencies_unstable.rosinstall'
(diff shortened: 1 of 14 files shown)


marip8:
Does the unstable build really add that much value? I guess its nice to know what the status of the build would be if you use the latest of everything, but the point of specifying dependencies by hash/version is to indicate the requirements for producing a successful and _intended_ build. I'm on the fence; I think my personal preference would be to do local builds with the latest changes on an individual developer basis and remove these extra files/CI jobs from the repository to minimize confusion

Levi-Armstrong:
> Does the unstable build really add that much value?

This is mainly for the maintainers so it obvious that upstream has changed and this repository has not been updated. I believe the ROS build farm does this for release packages for the same reason. This is to reduce the burden on the maintainer so they do not have to manually check if everything builds on master. 

Maybe we should make it a nightly build and not something that runs with every PR?

gavanderhoorn:
Does GHA already support `allowed-to-fail` builds, like Travis did?

If it does, then adding the *unstable* build as @Levi-Armstrong suggests would make sense to me: just mark it as `allowed-to-fail`.

Having information about breaking builds due to evolution in/of dependencies is valuable.

mpowelson:
I like it if it can be allowed to fail. It might make more sense for it to be a nightly build since it would break dependent on pull requests in other repos regardless of whether or not we are doing work in this repo. It seems that also fixes the allowed to fail issue since the only results of it failing would be a badge in the readme.

Levi-Armstrong:
Everyone, I have updated the settings for the master branch stating which CI are required to pass. I think if I were to use `allowed-to-fail` it would show up in the list as a pass even though it fails, but I will test to verify.

Levi-Armstrong:
It looks like the `continue-on-error` shows it passing which is not what we want.

PR #87, opened 2021-05-28
Update Descartes planner

marip8:
This PR makes various updates to the Descartes planner:
- Revise robot sampler
  - Don't do collision checking for redundant IK solutions
  - Add state cost based on distance from collision
- Leverage new state evaluator interface
- Add state evaluator to Descartes planning problem/profile

Depends on [this Tesseract PR](https://github.com/ros-industrial-consortium/tesseract/pull/601)


Levi-Armstrong:
@marip8 Can you rebase on the latest?

Levi-Armstrong:
@marip8 I have created issue #89 to track the progress of updating the planners to handle changes in how redundant solutions are calculated.

tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_default_plan_profile.h
```diff
@@ -63,20 +63,25 @@ class DescartesDefaultPlanProfile : public DescartesPlanProfile<FloatType>
...
+
   // Applied to sampled states
   bool enable_collision{ true };
   tesseract_collision::CollisionCheckConfig vertex_collision_check_config{ 0 };
 
   // Applied during edge evaluation
   bool enable_edge_collision{ false };
   tesseract_collision::CollisionCheckConfig edge_collision_check_config{ 0 };
+
+  bool use_redundant_joint_solutions{ false };
```
Levi-Armstrong, line 82: Doxygen. Also, this is a behavior change, because before it would use all redundant solutions? Should the default be true?

tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_default_plan_profile.h
```diff
@@ -63,20 +63,25 @@ class DescartesDefaultPlanProfile : public DescartesPlanProfile<FloatType>
 
 #ifndef SWIG
   DescartesEdgeEvaluatorAllocatorFn<FloatType> edge_evaluator{ nullptr };
+  DescartesStateEvaluatorAllocatorFn<FloatType> state_evaluator{ nullptr };
 
   // If not provided it adds a joint limit is valid function
   DescartesVertexEvaluatorAllocatorFn<FloatType> vertex_evaluator{ nullptr };
 #endif
 
+  bool allow_collision{ false };
```
Levi-Armstrong, line 72: Doxygen.

Levi-Armstrong (commented):
Overall looks good. I have a few comments and questions.

tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_default_plan_profile.h
marip8: That's probably a question of personal preference. To me it seems like more of a niche option that you would enable for special planning circumstances. When it is enabled, it can generate a huge amount of extra edges, adding to the existing problem of solving time and memory consumption.

Levi-Armstrong: I am good with the default being false. It would be good to add Doxygen explaining the behavior if true versus false.

marip8: Sounds good. I'll add the comments

tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_default_plan_profile.h
marip8: What do you think about removing this parameter? I believe we can get the same behavior from the inverse of `enable_collisions`

Levi-Armstrong:
@marip8 You good with me merging this?

marip8:
Yes, it's been tested and seems to be working as expected

johnwason:
Do we know why the "Unstable" CI is failing?

marip8:
The latest merged PR in `descartes_light` has some changes to bring it into conformance with clang-tidy. That PR happened after I finished changes for this PR, so I decided not to include them. #102 addresses these changes and should fix the unstable build

johnwason:
Can we get the unstable build problem figured out? I need it for the Python and conda packages.

Levi-Armstrong:
I am working on a PR now. Should have it wrapped up soon.

marip8:
> Can we get the unstable build problem figured out? I need it for the Python and conda packages.

There will be more changes happening to `descartes` in the near future, so updates to this repository will probably still lag those changes and this problem will likely persist. Is there a reason the python and conda packages depend on the latest version of `descartes` rather than the version specified in the `.rosinstall` file? The intent with the unstable build was more to understand if this repository was behind some of its dependencies, not to have it working all the time. I'm wondering if we should pursue some other strategy for not breaking these other Python and conda builds

johnwason:
The Python CI is currently failing, and it isn't clear how it should be configured. If I pin the tests to specific commits, it will be permanently behind the upstream and there will be no way to tell if changes to other repos break the Python wrappers.

Levi-Armstrong:
Tesseract unstable is now passing. I think we should add a unstable ci build to tesseract_python and update the current rosinstall with stable hashes.

PR #95, opened 2021-06-23
Add debugging/profiling code for Taskflows

mpowelson:
This adds some utilities for debugging/profiling Taskflows. Every Task generates a TaskInfo. These are returned to the user in the planning response. The idea is that the user stores these as their system operates - for example by serializing them to XML files. Then these utilities can be used offline to see what branches of the Taskflow were used, how long each Task usually takes, etc. 

Summary of changes

- Add boost serialization for TaskInfo
- Add elapsed time to TaskInfo base class
- Add TaskInfoProfiler that can be used to generate summary charts like below
- Add custom DOT dump function that will create a DOT file similar to the one in taskflow except adding the ability to include custom information. This currently only works for GraphTaskflows. Attached is part of an DOT file for a Taskflow I used to test it. Note the avg time and the frequency of each return value.

custom_dot_file

Summary output from new taskflow profiling example
```
|                               Task Name|          Occurances|        Min Time (s)|        Max Time (s)|        Avg Time (s)|       Return values|      Return value %|
|----------------------------------------------------------------------------------------------------------------------------------------------------------------------| 
|                                 TRAJOPT|                  35|            0.132803|            0.529303|            0.259503|              0, 1, |      31.42, 68.57, |
|                                    OMPL|                  35|          0.00487762|            0.104429|           0.0594033|              0, 1, |      31.42, 68.57, |
|                         Seed Min Length|                  35|           1.428e-06|            6.49e-06|         2.33274e-06|                 1, |             100.0, |
|                            Interpolator|                  35|         0.000599418|          0.00997626|          0.00314066|                 1, |             100.0, |
|       Iterative Spline Parameterization|                  24|         0.000884674|          0.00479578|          0.00213154|                 1, |             100.0, |
|       Discrete Contact Check Trajectory|                  24|         0.000562527|          0.00118135|          0.00075497|                 1, |             100.0, |
```

### DIFF
--- a/tesseract_process_managers/include/tesseract_process_managers/core/task_info.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/core/task_info.h
@@ -52,6 +52,7 @@ public:
   using Ptr = std::shared_ptr<TaskInfo>;
   using ConstPtr = std::shared_ptr<const TaskInfo>;
 
+  TaskInfo() = default;  // Required for serialziation
   TaskInfo(std::size_t unique_id, std::string name = "");
   virtual ~TaskInfo() = default;
   TaskInfo(const TaskInfo&) = default;
@@ -82,6 +83,14 @@ public:
   Instruction results_output{ NullInstruction() };
   /** @brief This is a clone of the environment at the beginning of the task (optionally set)*/
   tesseract_environment::Environment::ConstPtr environment{ nullptr };
+
+  bool operator==(const TaskInfo& rhs) const;
+  bool operator!=(const TaskInfo& rhs) const;
+
+private:
+  friend class boost::serialization::access;
+  template <class Archive>
+  void serialize(Archive& ar, const unsigned int version);  // NOLINT
 };
 
 /** @brief A threadsafe container for TaskInfos */
--- a/tesseract_process_managers/include/tesseract_process_managers/core/task_info.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/core/task_info.h
@@ -69,6 +69,9 @@ public:
 
   std::string message;
 
+  /** @brief elapsed_time Time spent in this task in seconds*/
+  double elapsed_time;
+
   /** @brief Instructions passed to task (optionally set) */
   Instruction instructions_input{ NullInstruction() };
   /** @brief Instructions after running the task (optionally set)*/
--- a/tesseract_process_managers/src/core/task_info.cpp
+++ b/tesseract_process_managers/src/core/task_info.cpp
@@ -30,6 +30,41 @@ namespace tesseract_planning
 {
 TaskInfo::TaskInfo(std::size_t unique_id, std::string name) : unique_id(unique_id), task_name(std::move(name)) {}
 
+bool TaskInfo::operator==(const TaskInfo& rhs) const
+{
+  static auto max_diff = static_cast<double>(std::numeric_limits<float>::epsilon());
+
+  bool equal = true;
+  equal &= return_value == rhs.return_value;
+  equal &= unique_id == rhs.unique_id;
+  equal &= task_name == rhs.task_name;
+  equal &= message == rhs.message;
+  equal &= tesseract_common::almostEqualRelativeAndAbs(elapsed_time, rhs.elapsed_time, max_diff);
+  equal &= instructions_input == rhs.instructions_input;
+  equal &= instructions_output == rhs.instructions_output;
+  equal &= results_input == rhs.results_input;
+  equal &= results_output == rhs.results_output;
+  return equal;
+}
+bool TaskInfo::operator!=(const TaskInfo& rhs) const { return !operator==(rhs); }
+
+template <class Archive>
+void TaskInfo::serialize(Archive& ar, const unsigned int /*version*/)
+{
+  ar& boost::serialization::make_nvp("return_value", return_value);
+  ar& boost::serialization::make_nvp("unique_id", unique_id);
+  ar& boost::serialization::make_nvp("task_name", task_name);
+  ar& boost::serialization::make_nvp("message", message);
+  ar& boost::serialization::make_nvp("elapsed_time", elapsed_time);
+
+  ar& boost::serialization::make_nvp("instructions_input", instructions_input);
+  ar& boost::serialization::make_nvp("instructions_output", instructions_output);
+  ar& boost::serialization::make_nvp("results_input", results_input);
+  ar& boost::serialization::make_nvp("results_output", results_output);
+
+  //  ar& boost::serialization::make_nvp("environment", environment);
+}
+
 void TaskInfoContainer::addTaskInfo(TaskInfo::ConstPtr task_info)
 {
   std::unique_lock<std::shared_mutex> lock(mutex_);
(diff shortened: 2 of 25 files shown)


mpowelson:
It looks like the Instructions in the TaskInfo are not getting serialized correctly. The XML just looks like below. @Levi-Armstrong am I doing something wrong? That appears to be the same way you are doing the waypoint in planInstruction

```
		<results_output>
			<instruction>
				<tx class_id_reference="4">
					<base></base>
					<impl></impl>
				</tx>
			</instruction>
		</results_output>
```

tesseract_process_managers/include/tesseract_process_managers/core/task_info.h
```diff
@@ -52,7 +52,7 @@ class TaskInfo
   using Ptr = std::shared_ptr<TaskInfo>;
   using ConstPtr = std::shared_ptr<const TaskInfo>;
 
-  TaskInfo(std::size_t unique_id, std::string name = "");
+  TaskInfo(std::size_t unique_id = 0, std::string name = "");
```
Levi-Armstrong, line 55: How is the unique_id normally previously set?

tesseract_process_managers/src/core/task_info.cpp
```diff
@@ -30,6 +30,46 @@ namespace tesseract_planning
...
+  ar& boost::serialization::make_nvp("results_input", results_input);
+  ar& boost::serialization::make_nvp("results_output", results_output);
+
+  //  ar& boost::serialization::make_nvp("environment", environment);
+}
+
+#include <boost/archive/xml_oarchive.hpp>
+#include <boost/archive/xml_iarchive.hpp>
+template void TaskInfo::serialize(boost::archive::xml_oarchive& ar, const unsigned int version);
+template void TaskInfo::serialize(boost::archive::xml_iarchive& ar, const unsigned int version);
```
Levi-Armstrong, line 71: Recommend moving this to the end outside the namespace

``` c++
#include <boost/archive/xml_oarchive.hpp>
#include <boost/archive/xml_iarchive.hpp>
template void tesseract_planning::TaskInfo::serialize(boost::archive::xml_oarchive& ar, const unsigned int version);
template void tesseract_planning::TaskInfo::serialize(boost::archive::xml_iarchive& ar, const unsigned int version);
```

Levi-Armstrong:
A general question. Did you try using any of the taskflow observer capabilities and ran into issues? 

Also, I recently switched the table in trajopt to have the name last column in the table which makes sure it is always formatted cleanly if the names ever get too long. It may be worth doing that here.

tesseract_process_managers/include/tesseract_process_managers/core/task_info.h
mpowelson: It is set at the beginning of conditionalProcess for each TaskGenerator. It should still be set. This is just for serialization.

mpowelson:
> A general question. Did you try using any of the taskflow observer capabilities and ran into issues?
> 
> Also, I recently switched the table in trajopt to have the name last column in the table which makes sure it is always formatted cleanly if the names ever get too long. It may be worth doing that here.

I haven't tried it. However, I wanted to be able to track what branches the taskflow went down across many attempts, so I didn't think the profiler included with taskflow did this. 

It's also worth noting that I'd eventually like to start logging the derived classes specific to the TaskGenerator so that we can look back at logs later and see why certain tasks returned the values that they did.

tesseract_process_managers/include/tesseract_process_managers/core/task_info.h
Levi-Armstrong: Oh, yea. I forgot for serialization it must have a default constructor. Maybe add documentation state that default value is only provided for serialization developer should provide it?
 
or something like this. 

``` c++
TaskInfo() = default // Required for serialziation
TaskInfo(std::size_t unique_id, std::string name = "");
```

tesseract_process_managers/src/core/task_info.cpp
Levi-Armstrong: Let me know if this solves the serialization issues.

mpowelson: That doesn't seem to have fixed it.

tesseract_process_managers/include/tesseract_process_managers/core/task_info.h
mpowelson: Yeah that's probably better. I changed it to that.

tesseract_process_managers/src/core/task_info.cpp
mpowelson: @Levi-Armstrong I added instructions to the serialization test, and that seems to work. However, it still doesn't work in my application. Maybe there is a bug in the saveInputs function? Or I'm doing something wrong?

```
    auto info_map = response->interface->getTaskInfoMap();
    std::vector<TaskInfo> task_info_vec;
    for (const auto& info : info_map)
      task_info_vec.push_back(*info.second);
    std::string task_info_string = Serialization::toArchiveStringXML<std::vector<TaskInfo>>(task_info_vec);
```

Levi-Armstrong: Have you confirmed that the derived class stored as the base class still serialized ok?

Levi-Armstrong: It looks like you are serializing a `std::vector` but did you include `#include <boost/serialization/vector.hpp>` in the file?

mpowelson: Yeah I did. The vector serializes fine. Serializing the base class seems to work. This passes and the xml file has instructions in it.

```
TEST(TesseractProcessManagersSerializeUnit, serializeDiscreteContactCheckTaskInfo)  // NOLINT
{
  auto task_info = std::make_shared<DiscreteContactCheckTaskInfo>(123456, "my task");
  task_info->elapsed_time = 123.456;
  task_info->message = "Test message";
  task_info->instructions_input = rasterExampleProgram();
  task_info->instructions_output = freespaceExampleProgramABB();
  task_info->results_input = freespaceExampleProgramIIWA();
  task_info->results_output = rasterDTExampleProgram();

  {  // Archive program to file
    std::string file_path = tesseract_common::getTempPath() + "serializeDiscreteContactCheckTaskInfo.xml";
    EXPECT_TRUE(Serialization::toArchiveFileXML<TaskInfo>(*task_info, file_path));

    TaskInfo ntask_info = Serialization::fromArchiveFileXML<TaskInfo>(file_path);
    EXPECT_TRUE(*task_info == ntask_info);
  }

  {  // Archive program to string
    std::string task_info_string = Serialization::toArchiveStringXML<TaskInfo>(*task_info, "task_info");
    EXPECT_FALSE(task_info_string.empty());

    TaskInfo ntask_info = Serialization::fromArchiveStringXML<TaskInfo>(task_info_string);
    EXPECT_TRUE(*task_info == ntask_info);
  }
}

```

mpowelson: @Levi-Armstrong I just realized that I was being stupid. There is a flag to turn on/off saving the Instructions and it defaults to off. We don't currently have a good way of setting that in the planning server, but regardless I think this is fine for now.

mpowelson:
I made those two changes you requested and discovered that the Instructions not serializing was not really an issue.

Levi-Armstrong (approved):

PR #114, opened 2021-09-13
Motion Planner Package Reorganization

marip8:
This PR moves the implementations of the Descartes, OMPL, Trajopt, and Trajopt IFOPT motion planners into sub-directories that can be selectively built using CMake variables in order to improve build flexibility based on what dependencies a user is willing to support


Levi-Armstrong:
@marip8 Looks good. One question should the tests be moved into the respected directories core, descartes, ompl, trajopt, etc?

marip8:
Currently I put CMake `if` statements around the tests based on which planners they need to link against, and this seemed be straightforward and work well. Some of the tests require multiple planners (i.e. OMPL and Trajopt, Descartes and Trajopt, etc.) so I think it makes sense to keep the single test directory. Also, it might be a hassle to create a new test sub-directory in each of the planner sub-directories

Levi-Armstrong:
The only thing left is to figure out why the windows build is failing.

marip8:
> /MANIFESTFILE:tesseract_motion_planners_descartes.dll.manifest" failed (exit code 1181) with the following output:
2021-09-14T18:57:48.6222298Z LINK : fatal error LNK1181: cannot open input file 'LinearMath.lib

Seems to be failing on the `descartes` library due to a library that looks to be part of `bullet`

PR #122, opened 2021-10-15
CI Update

marip8:
This PR updates the CI configuration to use the `docker` images produced by the `tesseract` CI builds as the base image and workspace underlay. This should increase the speed of builds since `tesseract` and its dependencies won't need to be built from source every time. 

This change will also push built images of this project to the Github container registry.


marip8:
Looks like we need to bump the version of `ros-industrial-cmake-boilerplate` up from 0.2.8 in the `tesseract` `docker` image to get this to build correctly. Otherwise it seems to be configured correctly

PR #136, opened 2021-11-18
Fix ability to use same task with different parameters adding namespaces to the profile dictionary

Levi-Armstrong:
This was possible in the motion planners prior to moving the profile dictionary to the planning request. After moving the profile dictionary to the planning request it was no longer possible to using the same task in a taskflow and have different parameters. This PR updates the profile dictionary so profiles are under a namespace which in the case of taskflow the namespace is the task names. This allows you to use the same task with a different name and associated profiles under this namespace in the profile dictionary.


Levi-Armstrong:
Also I have tested this with tesseact_ros_examples and everything runs as it did before this change.

marip8:
Can you provide a concrete example (description and/or task flow diagram) of why this is necessary? I don't completely understand the use-case from the description

Levi-Armstrong:
> Can you provide a concrete example (description and/or task flow diagram) of why this is necessary? I don't completely understand the use-case from the description

I don't have a diagram, but for example if you had a pipeline where you started with  TrajOpt (the profile where collision is  a cost) and then after success you run through TrajOpt again (with profile where collision is a constraint). For a given composite you have a single profile name only allowing a single entry in the profile dictionary. This was possible to do before because the profiles were directly added to the motion planner, but know they are using the profile dictionary. Since it is a single profile name there is no way to distinguishes between the two TrajOpt tasks. Adding such that the profiles are namepsaced in the profile dictionary then each trajopt task would have its own namespace with a single profile corresponding to the profile name in the composite instruction. This provides more flexibility for error pipeline where on failure you could have it run the same task with different parameters. Hope this helps.

PR #138, opened 2021-11-23
WIP: New Profile Interfaces and OMPL Planner Refactor

marip8:
This PR introduces new interfaces for waypoint, composite, and planner profiles to support a refactor of the motion planners. The OMPL motion planner was also updated to utilize the new profile definitions.

# To-do
- [ ] Understand why JointStartCartesianGoal unit test fails (currently produces only 1 valid OMPL goal state)
- [ ] Cast `data` object to `ompl::base::SimpleSetup` if it is not a `nullptr`
- [ ] Decide which instructions to extract planner and composite profiles from


tesseract_motion_planners/ompl/src/profile/ompl_composite_profile_rvss.cpp
```diff
@@ -0,0 +1,360 @@
...
+      for (const auto& contact_vec : contact_map_vec[i])
+        for (const auto& contact : contact_vec.second)
+          CONSOLE_BRIDGE_logError(("Solution: " + std::to_string(i) + "  Links: " + contact.link_names[0] + ", " +
+                                   contact.link_names[1] + "  Distance: " + std::to_string(contact.distance))
+                                      .c_str());
+    throw std::runtime_error("In OMPLDefaultPlanProfile: All goal states are either in collision or outside limits");
+  }
+
+  return valid_solutions;
+}
```
marip8, line 125: It takes a surprisingly large amount of code to perform collision-aware IK. We should consider making this a capability of the `KinematicGroup` or a utility function somewhere so we don't have to continually rewrite this

Levi-Armstrong: Is the same code used anywhere else? I believe this is the only place where this is being done. 

For others that may read this it is doing more than just collision-aware IK. It checks and enforces bounds, calculates redundant solutions, along with print debug information if no valid solution is found.

marip8: It's not necessarily being used elsewhere yet, but having this collision-aware IK capability is really valuable to a lot of other applications. In those cases, someone would have to reimplement this function because it doesn't formally exist in `tesseract`. My suggestion would be to move this somewhere like the `KinematicsGroup` to expose it to other applications because it is generic and valuable

Levi-Armstrong: I agree, the only thing with adding it to the kinematic group is it would now make tesseract_kinematics depend on tesseract_collision which I am not opposed to but wanted to mention it.

marip8: Yeah, I thought about that as well. Would it make sense to move the high-level manipulator classes (`KinematicGroup` and possibly `JointGroup` for consistency) to `tesseract_environment` where they could have access to collision checkers? In my mind the value of the manipulator classes is that they unite the various underlying pieces (IK, FK, enforcing joint limits, getting redundant solutions, checking collision, etc.) for moving a robot into a high-level easily usable class. `tesseract_environment` seems to be the container for that high-level capability

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -104,439 +95,237 @@ bool OMPLMotionPlanner::terminate()
...
 
-  /// @todo: Need to expand this to support multiple motion plans leveraging taskflow
-  for (auto& p : problem)
+  // Get the planner profile
+  OMPLPlannerParameters params;
+  try
   {
-    auto parallel_plan = std::make_shared<ompl::tools::ParallelPlan>(p->simple_setup->getProblemDefinition());
+    const std::string profile_name =
+        getProfileString(name_, request.instructions.getProfile(), request.plan_profile_remapping);
```
marip8, line 123: Where should the planner profile come from? Right now I'm extracting it from the request instructions composite it seems like that should be where the composite profile comes from

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -104,439 +95,237 @@ bool OMPLMotionPlanner::terminate()
...
+  }
 
-std::vector<OMPLProblem::Ptr> OMPLMotionPlanner::createProblems(const PlannerRequest& request) const
-{
-  std::vector<OMPLProblem::Ptr> problem;
-  tesseract_kinematics::JointGroup::Ptr manip;
+  // Initialize the output by copying the input program and flattening it
+  response.results = request.seed;
+  std::vector<std::reference_wrapper<Instruction>> results_flattened =
+      flattenProgramToPattern(response.results, request.instructions);
```
marip8, line 275: Is there a better/safer way to fill out the output trajectory? It's problematic to set the response equal to the seed and then overwrite the waypoints. I just fixed an issue where I was accidentally skipping the overwrite of the 0th element of the trajectory composite, but nothing failed and it wasn't obvious because there was already a default waypoint in that spot from the seed.

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -104,439 +95,272 @@ bool OMPLMotionPlanner::terminate()
...
-    // Enforce limits
-    for (Eigen::Index i = 0; i < trajectory.rows(); i++)
-      tesseract_common::enforcePositionLimits(trajectory.row(i), p->manip->getLimits().joint_limits);
+  // Element 0 of the flattened results is a start instruction that should contain the first trajectory state
+  auto& start_instruction = results_flattened.at(0).get().as<MoveInstruction>();
 
-    assert(checkStartState(p->simple_setup->getProblemDefinition(), trajectory.row(0), p->extractor));
-    assert(checkGoalState(p->simple_setup->getProblemDefinition(), trajectory.bottomRows(1).transpose(), p->extractor));
+  // Element 1 of the flattened results is a composite instruction containing the reset of the trajectory
+  auto& composite = results_flattened.at(1).get().as<CompositeInstruction>();
```
marip8, line 308: It seems confusing to have a separate move instruction for the start of the trajectory and to have the rest of the trajectory inside a composite instruction, which itself has a start instruction placeholder which happens to be a `NullInstruction`. Is there a reason it was set up like this in the previous version? Would it be better to put the first trajectory waypoint into the start instruction of the composite with the rest of the trajectory?

marip8:
@Levi-Armstrong I think this is ready for review. The unit tests pass (locally), and I'm working on visualizing the freespace example to confirm proper behavior

Levi-Armstrong:
It looks like several of the process planner units are now failing.

marip8:
I think the process planner unit tests are failing because no default profiles were added to the new profile dictionary maps for the OMPL planner. I'll look into it. Strange that the Windows build succeeds. Does it not run unit tests?

Levi-Armstrong:
Yea it does not run the unit test on the windows build. It just makes sure everything build. A few were segfaulting which I believe it's due to trying to write to temp directory.

marip8:
I resolved the issue with the unit tests, and I was able to successfully run the examples in this repository (and `tesseract_ros_examples` with a few minor changes). It looks like code coverage fails, presumably because of the few functions that I added to the profile dictionary to support this. I could add a unit test for this capability in this PR, but I think the idea is that we would eventually replace the existing profile dictionary with the new implementation once all the planners are updated. I would propose holding off on making those unit tests until we are ready to finish the profile dictionary

PR #141, opened 2021-11-29
Update dependency tags

Levi-Armstrong:


marip8:
@Levi-Armstrong can you push tags to this repository to correspond with the tags in `tesseract`? This repository is currently at 0.6.5 but `tesseract` is now at 0.6.8. I'm not sure which commits that were added recently here correspond to which tags in `tesseract`

Levi-Armstrong:
I didn't think they need to be the same number; is that what you are requesting? If you checkout tesseract planning it should have the version of tesseract required. I will update the readme to list the version of dependencies so it is easier to find

marip8:
> I didn't think they need to be the same number; is that what you are requesting?

Yes. I thought the strategy we had discussed discussed was to tag this repository the same as the corresponding version of `tesseract`. So if I pulled `tesseract-0.6.5` I would know that I could build this repository at `0.6.5` or with any commit up to the next higher tag. This seems to be what other projects, like `boost`, do that separate their repositories.

The required version of `tesseract` can be documented in the `.rosinstall` file and the CI files, but the first place I would look for a specific version of a repository is in the tags. Plus, it makes it easy for projects that depend on this repository to pull at the correct version. Documenting the dependency in the README is helpful, but we'll need to be diligent about keeping it up to date when we change CI. It seems like it would be easier to document the dependency strategy in the README (i.e. use the tag in this repository that corresponds to `tesseract`) and push tags, rather than updating the README every time the dependency changes.

Levi-Armstrong:
I think it will be challenging to keep them and other like tesseract_ros, tesseract_gui, tesseract_ros_workcell the same version. I imagine that as things progress tesseract will be changing very little and the other will be changing more frequently and we may want to tag frequently which would require this to propagate down which is a pain from a maintenance perspective. For example if I want to tag tesseract_ros, I would need to tag tesseract and wait for the docker to be uploaded. Then tag tesseract_planning and wait for the dockers to be uploaded and then tag tesseract_ros. If we allow them to be different but document then you would only need to update tesseract_ros.

gavanderhoorn:
It's more maintenance overhead, but what about following [semver](https://semver.org/)?

So any `x.y` tag from `tesseract` should be compatible with any `x` tag in any of the other repositories (so `1.x` is compatible with all other `1.m.n`). Newer or older major versions are not guaranteed to work.

If there is no such compatibility (because something really needed to be "broken" to be fixed), increase the major version to clearly signal this. Minor version nrs increase if changes are bw compatible.

If backporting functionality or bug fixes would be desirable, tags don't work so well any more, and you'd probably need to consider using a more elaborate branching structure/workflow.

marip8:
For the record, I'm not asking that we always keep this repository up-to-date with `tesseract`, only that the existing tags represent the dependency accurately in the sense described above (i.e. any commit between 0.6.5 (inclusive) - 0.6.6 (exclusive) build with `tesseract-0.6.5`). At the time I posted, the last tag was `0.6.5` but some commits ahead required advanced versions of `tesseract`, including the master branch which needs `tesseract-0.6.8`. For now, I'm just asking that as we update this repository to be compatible with later versions of `tesseract` we also tag as those changes are merged to indicate the dependency clearly.

> It's more maintenance overhead, but what about following semver?

We haven't been following `semver` very well up to this point since the last 5 or 6 changes have been API-breaking but were documented as minor changes, but it's probably a good idea to start using it. Eventually when `tesseract` stabilizes a little more, I could see us making changes to this repository that we would want to tag separately, and `semver` would facilitate that better like you describe

PR #149, opened 2021-12-13
Only check kinematics if built in debug

Levi-Armstrong:
This also includes a fix for the process planners that perform a global plan first. It would hard code the raster start and end position to the results of the global plan even if they were cartesian waypoints. This fixes it such that the start and end position are based on the input instruction so if it is a cartesian waypoint it is still able to improve the results during the raster planning.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -18,6 +18,7 @@ Checks: >
     -cppcoreguidelines-non-private-member-variables-in-classes,
     misc-*,
     -misc-non-private-member-variables-in-classes,
+    -misc-no-recursion,
     modernize-*,
     -modernize-use-trailing-return-type,
     -modernize-use-nodiscard,
@@ -26,7 +27,8 @@ Checks: >
     -readability-braces-around-statements,
     -readability-named-parameter,
     -readability-magic-numbers,
-    -readability-isolate-declaration
+    -readability-isolate-declaration,
+    -readability-function-cognitive-complexity
 WarningsAsErrors: >
     -*,
     clang-diagnostic-*,
@@ -46,6 +48,7 @@ WarningsAsErrors: >
     -cppcoreguidelines-non-private-member-variables-in-classes,
     misc-*,
     -misc-non-private-member-variables-in-classes,
+    -misc-no-recursion,
     modernize-*,
     -modernize-use-trailing-return-type,
     -modernize-use-nodiscard,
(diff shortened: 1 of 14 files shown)


Levi-Armstrong:
@marrts I added @marip8, but if you are able to review instead please do.

marip8:
I'm not a fan of enabling/disabling code based on build types. I personally almost always build in debug so I can investigate problems when they inevitably arise, and I also don't want to be forced to check kinematics for a manipulator group that I have already verified.

I would suggest removing the call to `checkKinematics` from the planners altogether. If we want to check the kinematics of a group, it should probably be done in the `Environment` when the group is created or modified since tasks other than planning would also be affected.

If we still want to include this functionality in the planning pipeline, it seems like the most flexible way would be to make a standalone task to do it. If a user needs to add this capability to their planning pipeline they could just create a custom graph task flow with this task as the first step.

Levi-Armstrong:
> I'm not a fan of enabling/disabling code based on build types. I personally almost always build in debug so I can investigate problems when they inevitably arise, and I also don't want to be forced to check kinematics for a manipulator group that I have already verified.

The purpose of things like this are to help the developer, by alerting you of errors that should never occur. This is exactly what every assert does does under the hood used throughout the code base. The only reason I did not use an assert here is because some kinematics may slightly fail due to the tolerance. I think it wold be good to move the check as you suggested to the environment so it is not in the planning pipeline which would cause this function to be called multiple time in debug, but I do not think removing it completely is a good idea. This check helped find mutliple issues like the multi-threading issues in KDL and errors in OPW kinematics. I will move it to the getKinematicsGroup in the environment. This way it would only every get called once for a given manipulator on creation.

marip8:
> This check helped find multiple issues like the multi-threading issues in KDL and errors in OPW kinematics.

I understand the value of the `checkKinematics` function, but it doesn't seem like making the check in the planners is the right place for it. I think the right solution is to check the kinematics on creation or modification of a group, as proposed [here](https://github.com/tesseract-robotics/tesseract/pull/682)

> I will move it to the getKinematicsGroup in the environment. This way it would only every get called once for a given manipulator on creation.

Once [this `tesseract` PR](https://github.com/tesseract-robotics/tesseract/pull/682) gets merged, it should be the case that the kinematics of all created groups will have been checked by the environment on creation. Wouldn't the same check in the planners be redundant at that point?

Levi-Armstrong:
> I understand the value of the `checkKinematics` function, but it doesn't seem like making the check in the planners is the right place for it. I think the right solution is to check the kinematics on creation or modification of a group, as proposed [here](https://github.com/tesseract-robotics/tesseract/pull/682)

I agree.

> Once this tesseract PR gets merged, it should be the case that the kinematics of all created groups will have been checked by the environment on creation. Wouldn't the same check in the planners be redundant at that point?

I have removed it from the planners.

PR #156, opened 2022-01-07
Revised Planner Profile Interfaces

marip8:
This PR provides new interfaces for motion planner waypoint, composite, and planner level profiles. Each profile produces a type-erased, planner-specific data object from a relevant input (an instruction for the waypoint profile, a composite instruction for the composite profile). The individual planners can look up profiles in the profile dictionary by a combination of namespace and name, and can cast them to the required type. Since the objects returned by the profiles are `std::any` type-erasers, any issues with incorrect casts will result in an exception that can be handled.

The addition of new planner profile interfaces supports an on-going larger refactor and simplification of the various motion planners class (started in #138). Additionally, the new profiles will be easier to store and retrieve in the profile dictionary since they will no longer be stored by class `typeid`.

This PR adds three new maps in the profile dictionary to hold these waypoints. Currently these maps are public members of the profile dictionary for convenience and to minimize API breakage. As motion planners are refactored to use these profile interfaces, the planners will retrieve profiles from these maps. Ultimately, the map storing profiles by `typeid` will be removed, the three maps holding the new profiles will be made private, and the existing profile dictionary API will be revised to access the new profile maps.

### DIFF
--- a/tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
+++ b/tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
@@ -36,6 +36,9 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <shared_mutex>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
+#include <tesseract_command_language/composite_instruction.h>
+#include <tesseract_environment/environment.h>
+
 #ifdef SWIG
 %shared_ptr(tesseract_planning::ProfileDictionary)
 #endif  // SWIG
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
This does not break existing functionality correct? If it does not, then I am good with merging as is for now and updating after motion planning has been fully refactored as you suggested.

marip8:
> This does not break existing functionality correct? If it does not, then I am good with merging as is for now and updating after motion planning has been fully refactored as you suggested.

Correct. It only adds additional code to the profile dictionary that will be used in the future when the planners are refactored.

Levi-Armstrong (approved):

marip8:
Closing; superceded by #267

PR #157, opened 2022-01-07
Instruction Start State Update

marip8:
Currently the start state of most composite instructions (especially nested composite instructions) are never changed from `NullInstructions`. The reason for this is that usually the start instruction of one composite usually duplicates the last instruction of the previous composite, such as in the case of a robot trajectory. To avoid duplicating waypoints, the start instructions of most subsequent composites are left empty. However, this can be problematic when trying to use a nested composite instruction as a seed for another instruction or when using the composite instruction on its own. It also makes the composite instruction confusing to interpret since it is missing its first waypoint.

This PR updates the instruction `flatten` functions to ignore start instructions from composites that are not the first in a container. This retains the current behavior for the assembly of trajectories from nested composite instructions. This PR also changes the simple planner utility `getInterpolatedComposite` to put the first state of the interpolation as the start instruction of the composite, such that the entire interpolation is captured in the composite instruction.

This change is required for the OMPL planner refactor in #138

### DIFF
--- a/tesseract_command_language/include/tesseract_command_language/utils/get_instruction_utils.h
+++ b/tesseract_command_language/include/tesseract_command_language/utils/get_instruction_utils.h
@@ -35,6 +35,7 @@ namespace tesseract_planning
 {
 /**
  * @brief Get the first Instruction in a Composite Instruction that is identified by the filter
+ * @details This does not consider the start instruction of nested composite instructions
  * @param composite_instruction Composite Instruction to search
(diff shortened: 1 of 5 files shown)


Levi-Armstrong:
I don't think this will break things but are have you ran the tesseract_ros_examples to make sure they still work?

marip8:
Closing; this seems to have been invalidated by #268

PR #160, opened 2022-01-11
Make ProcessPlanningServer::run const

mpowelson:
The only thing keeping run from being const is that executor_->run is non-const even though it appears to be threadsafe.


Levi-Armstrong (approved):
Once you fix the CI issue it should be ready to merge.

PR #162, opened 2022-01-13
Add user data to plan and move instruction

Levi-Armstrong:
This adds the ability to store user data in the plan and move instruction.

### DIFF
--- a/tesseract_command_language/include/tesseract_command_language/move_instruction.h
+++ b/tesseract_command_language/include/tesseract_command_language/move_instruction.h
@@ -39,6 +39,7 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 #include <tesseract_command_language/constants.h>
 #include <tesseract_command_language/profile_dictionary.h>
 #include <tesseract_command_language/types.h>
+#include <tesseract_common/any.h>
 
 namespace tesseract_planning
 {
@@ -138,6 +139,12 @@ public:
 
   bool operator==(const MoveInstruction& rhs) const;
   bool operator!=(const MoveInstruction& rhs) const;
+
+  /**
+   * @brief Store user data
+   * @brief The object must be boost serializable
+   */
+  tesseract_common::Any user_data;
 
 private:
   MoveInstructionType move_type_{ MoveInstructionType::START };
--- a/tesseract_command_language/include/tesseract_command_language/plan_instruction.h
+++ b/tesseract_command_language/include/tesseract_command_language/plan_instruction.h
@@ -38,6 +38,7 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 #include <tesseract_command_language/constants.h>
 #include <tesseract_command_language/profile_dictionary.h>
 #include <tesseract_command_language/types.h>
+#include <tesseract_common/any.h>
 
 namespace tesseract_planning
 {
@@ -130,6 +131,12 @@ public:
 
   bool operator==(const PlanInstruction& rhs) const;
   bool operator!=(const PlanInstruction& rhs) const;
+
+  /**
+   * @brief Store user data
+   * @brief The object must be boost serializable
+   */
+  tesseract_common::Any user_data;
 
 private:
   PlanInstructionType plan_type_{ PlanInstructionType::START };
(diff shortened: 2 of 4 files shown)


marip8:
Why is this necessary? `Waypoint` is already a type eraser. If a user wants to embed custom data, they should just make their own waypoint definition with the data they require, inheriting from the Cartesian, joint or state waypoint for convenience if necessary. A major point of making `Waypoint` a type-eraser was that it could remain simple while allowing users to easily customize their own implementations. Adding a type-eraser to the `Waypoint` type-eraser seems unnecessarily redundant

Levi-Armstrong:
> Why is this necessary? `Waypoint` is already a type eraser. If a user wants to embed custom data, they should just make their own waypoint definition with the data they require, inheriting from the Cartesian, joint or state waypoint for convenience if necessary. A major point of making `Waypoint` a type-eraser was that it could remain simple while allowing users to easily customize their own implementations. Adding a type-eraser to the `Waypoint` type-eraser seems unnecessarily redundant

It did not work as I expected adding it the type erasure class. I have updated the PR to only add user data to the plan and move instruction.

Levi-Armstrong:
I did some testing and we could add a new required method to the type erasure to enable this

``` c++
class A
{
public:
    double a;
    virtual bool isConvertable(std::type_index id) const
    {
      return (id == std::type_index(typeid(A)));
    }
};

class B : public A
{
public:
    double b;
    bool isConvertable(std::type_index id) const override
    {
        if (id == std::type_index(typeid(B)))
            return true;
        
        return A::isConvertable(id);
    }
};
```

marip8:
Can you add some comments to the lines in your code block to explain what doesn't work? I'm assuming only the `isCartesianWaypoint` doesn't work because it specifically checks if the `typeid` matches that particular class.

If that's the problem, I would suggest changing the helper functions to attempt a cast/catch the exception and return the status of the cast 

```c++
template<typename WaypointT>  // template so same function can be used for any waypoint types
bool isWaypoint(const Waypoint& waypoint)
{
  try
  {
    waypoint.as<WaypointT>();
    return true;
  }
  catch(const std::bad_cast&)
  {
    return false;
  }
}
```

I would assume the cast would work because `DerivedCartesianWaypoint` can be interpreted as a `CartesianWaypoint` since it derives from it

Levi-Armstrong:
The issue is not in the `isWaypoint` method but in the type erasure itself. Current the type erasure `as` methods do the same thing std::any does with checking the type index and they must exactly match so casting to base type is not allowed. It could be change to not check the type index but I would prefer not to do that because it has to do a static_cast which will make it hard to track down bad casts. I think the only other option is to add the required method `isConvertable`.

``` c++
Waypoint dcw = DerivedCartesianWaypoint(...);
dcw.as<CartesianWaypoint>(dcw);  // Would like this to be a valid cast, but fails because the type_index do not match
dcw.as<DerivedCartesianWaypoint>(dcw);  // returns true
dcw.as<JointWaypoint>(dcw);  // returns false;
```

marip8:
> The issue is not in the isWaypoint method but in the type erasure itself. Current the type erasure as methods do the same thing std::any does with checking the type index and they must exactly match so casting to base type is not allowed.

Just when we thought type erasers were the answers to all our problems... That's unfortunate; I think the `isConvertable` method would be a decent solution that isn't too invasive

Levi-Armstrong:
> > The issue is not in the isWaypoint method but in the type erasure itself. Current the type erasure as methods do the same thing std::any does with checking the type index and they must exactly match so casting to base type is not allowed.
> 
> Just when we thought type erasers were the answers to all our problems... That's unfortunate; I think the `isConvertable` method would be a decent solution that isn't too invasive

Yea, I am not sure we are using them as they should be. I believe the Plan and Move Instruction along with the maypoint types should also be type erasure because they are interfaces for the motion planning. Then user could assign there own types to these type erasure which then get stored as instruction. Then planning would just cast the instruction to the type erased plan instruction and use the interface methods and then user could cast the plan instruction to the known type. Then the motion planning framework does not have to cast other than the instruction to the other type erased interfaces.This might be the right solution, because then we do not need to have this isConvertable.

Levi-Armstrong:
This will no longer be needed with the upcoming type erasure changes.

PR #176, opened 2022-03-25
Add ctest calls to windows_noetic_build.yml

johnwason:
It appears that tests were not being run correctly in windows_noetic_build.yml. I explicitly ran ctest on the packages and got a lot of unit test failures. This PR adds explicit ctest calls to the workflow script.


Levi-Armstrong:
Yea, I intentional disabled the tests on windows because I do not have a windows machine to go through and address the issue.

johnwason:
Hmm the tests aren't running correctly. I will see why.

You can use something like virtualbox to run windows on your linux machine. Windows doesn't require a license key these days, they just have a nag prompt to ask you to register. The performance is pretty close to a normal machine if you enable kvm.

PR #178, opened 2022-03-27
Fix windows CI build

Levi-Armstrong:


Levi-Armstrong:
@mpowelson The issue with the windows build was the its rosinstall did not get updated. Not sure why it gives cryptic error messages.

PR #184, opened 2022-04-07
Add isValidCast to instruction and waypoint type erasure

Levi-Armstrong:
@marip8  and @mpowelson This will enable to use derived types for waypoint types ans instruction types. I have the waypoint finished along with a unit test which was failing before the change and now passes. I am working on instructions now. Let me know what you think.


Levi-Armstrong:
This will not be need with the upcoming type erasure changes.

PR #185, opened 2022-04-08
Fix logic in FixStateBounds for case ALL

mpowelson:
This fixes a logic mistake in FixStateBounds that was causing it to always clamp the limits


Levi-Armstrong (approved):

Levi-Armstrong:
The unstable threw and exception not sure if it is related to this package. Going to restart and see if it happens again.

PR #197, opened 2022-05-23
Make process planning server threadsafe and support multiple executors

Levi-Armstrong:
This now allows for user defined executors which is selected as part of the process planning request. This essentially allows you to define different thread pools instead of just have a single one.

### DIFF
--- a/tesseract_process_managers/include/tesseract_process_managers/core/process_planning_request.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/core/process_planning_request.h
@@ -39,9 +39,15 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_planning
 {
+/** @brief The name of the primary executor (thread pool) */
+static const std::string PRIMARY_EXECUTOR_NAME = "primary";
+
 class ProcessPlanningRequest
 {
 public:
+  /** @brief The name of the executor to use */
+  std::string executor{ PRIMARY_EXECUTOR_NAME };
+
   /** @brief The name of the Process Pipeline (aka. Taskflow) to use */
   std::string name;
 
--- a/tesseract_process_managers/include/tesseract_process_managers/core/process_planning_request.h
+++ b/tesseract_process_managers/include/tesseract_process_managers/core/process_planning_request.h
@@ -46,7 +46,7 @@ class ProcessPlanningRequest
 {
 public:
   /** @brief The name of the executor to use */
-  std::string executor{ PRIMARY_EXECUTOR_NAME };
+  std::string executor_name{ PRIMARY_EXECUTOR_NAME };
 
   /** @brief The name of the Process Pipeline (aka. Taskflow) to use */
   std::string name;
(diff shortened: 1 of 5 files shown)


tesseract_process_managers/include/tesseract_process_managers/core/process_planning_server.h
```diff
@@ -87,6 +90,27 @@ class ProcessPlanningServer
   ProcessPlanningServer& operator=(ProcessPlanningServer&&) = default;
 #endif  // SWIG
 
+  /**
+   * @brief Add a executors (thread pool) under the provided name
+   * @brief This creates a taskflow executor with the provided number of threads
+   * @param name The name of the thread pool
+   * @param n The number of threads
+   */
+  void addExecutor(const std::string& name, size_t n);
```
mpowelson, line 99: Should we just pass in the executor? Then you could share executors between servers if you wanted.

tesseract_process_managers/include/tesseract_process_managers/core/process_planning_request.h
```diff
@@ -39,9 +39,15 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
 namespace tesseract_planning
 {
+/** @brief The name of the primary executor (thread pool) */
+static const std::string PRIMARY_EXECUTOR_NAME = "primary";
+
 class ProcessPlanningRequest
 {
 public:
+  /** @brief The name of the executor to use */
+  std::string executor{ PRIMARY_EXECUTOR_NAME };
```
mpowelson, line 49: Maybe executor_name? I was a bit confused when I saw in the cpp that you were doing executor==rhs.executor since I didn't realize it was actually the name not the executor itself.

tesseract_process_managers/include/tesseract_process_managers/core/process_planning_server.h
```diff
@@ -163,12 +197,36 @@ class ProcessPlanningServer
...
+  /**
+   * @brief Queries the number of running tasks at the time of this call
+   * When a taskflow is submitted to an executor, a topology is created to store
+   * runtime metadata of the running taskflow.
+   * @param name The name of the executor to get task count for
+   */
+  long getTaskCount(const std::string& name = PRIMARY_EXECUTOR_NAME) const;
+
 protected:
+  mutable std::shared_mutex mutex_;
```
mpowelson, line 215: Maybe a more descriptive name? It looks like this mutex is specifically for the executor map

tesseract_process_managers/include/tesseract_process_managers/core/process_planning_server.h
Levi-Armstrong: Create idea. I added another method to support this.

tesseract_process_managers/include/tesseract_process_managers/core/process_planning_server.h
Levi-Armstrong: It is primarily used for all three maps withing the class. 
``` c++
/** @brief A map of executors which are thread safe */
  std::unordered_map<std::string, std::shared_ptr<tf::Executor>> executors_;

  /** @brief A map of profile observers */
  std::unordered_map<std::string, std::shared_ptr<tf::TFProfObserver>> profile_observers_;

  /** @brief A map of process planners */
  std::unordered_map<std::string, TaskflowGenerator::UPtr> process_planners_;
```

mpowelson: Ok. Well this obviously doesn't really matter, so I think it's fine.

Issue #201, opened 2022-06-05
Excess Collision Checking in Continuous Collision Checking


marrts:
https://github.com/tesseract-robotics/tesseract_planning/blob/9d08e6a0a995e422a858c9fb254306dc47253514/tesseract_motion_planners/core/src/core/utils.cpp#L317-L374
This appears to iterate over every step every time any adjacent step has a distance less than the LVS, potentially causing an n^2 order of magnitude increase in computation. It should just be doing collision checking between the already made swp0 and swp1.

Levi-Armstrong:
WOW, good find.

PR #202, opened 2022-06-05
Fix issue #201 excess collision checking in contactCheckProgram with ContinuousContactManager

Levi-Armstrong:


Levi-Armstrong:
@marrts Do you mind reviewing?

marrts:
This looks good, exactly what I figured it was supposed to be.

PR #203, opened 2022-06-14
Update to use new type erasure interface

Levi-Armstrong:
- This updated to leverage the new type erasure interface to reduce code duplication. 
- Leverage Boost Concept check to confirm type interface at compile time

@mpowelson, @marip8 and @marrts  Need help with defining naming convention for type erasure. Currently it is not obvious based on the type to know if it is a type erasure or not. I noticed when reviewing adobe example they prefix the type with Poly. Let me know what you think or if you have alternative naming convention. 

Current Type Erasures
- Instruction
- Waypoint

Future Type Erasures (This is to support custom implementations)
- MoveInstruction
- CartesianWaypoint
- JointWaypoint
- StateWaypoint

Example suffix with Poly
- MoveInstructionPoly, This is the type erasure
  - MoveInstruction, Default implementation

Example prefix with Poly
- PolyMoveInstruction, This is the type erasure
  - MoveInstruction, Default implementation

### DIFF
--- a/.github/workflows/dependencies_windows_build.rosinstall
+++ b/.github/workflows/dependencies_windows_build.rosinstall
@@ -9,7 +9,7 @@
 - git:
     local-name: tesseract
     uri: https://github.com/ros-industrial-consortium/tesseract.git
-    version: 0.9.9
+    version: 0.9.10
 - git:
     local-name: trajopt
     uri: https://github.com/ros-industrial-consortium/trajopt_ros.git
--- a/tesseract_command_language/include/tesseract_command_language/cartesian_waypoint.h
+++ b/tesseract_command_language/include/tesseract_command_language/cartesian_waypoint.h
@@ -223,7 +223,7 @@ private:
 #ifdef SWIG
 %tesseract_command_language_add_waypoint_type(CartesianWaypoint)
 #else
-TESSERACT_WAYPOINT_EXPORT_KEY(tesseract_planning::CartesianWaypoint);
+TESSERACT_WAYPOINT_EXPORT_KEY(tesseract_planning, CartesianWaypoint);
 #endif  // SWIG
 
 #endif  // TESSERACT_COMMAND_LANGUAGE_CARTESIAN_WAYPOINT_H
--- a/tesseract_command_language/include/tesseract_command_language/composite_instruction.h
+++ b/tesseract_command_language/include/tesseract_command_language/composite_instruction.h
@@ -286,7 +286,7 @@ private:
 #ifdef SWIG
 %tesseract_command_language_add_instruction_type(CompositeInstruction)
 #else
-TESSERACT_INSTRUCTION_EXPORT_KEY(tesseract_planning::CompositeInstruction);
+TESSERACT_INSTRUCTION_EXPORT_KEY(tesseract_planning, CompositeInstruction);
 #endif  // SWIG
 
 #endif  // TESSERACT_COMMAND_LANGUAGE_COMPOSITE_INSTRUCTION_H
--- a/tesseract_command_language/include/tesseract_command_language/core/instruction.h
+++ b/tesseract_command_language/include/tesseract_command_language/core/instruction.h
@@ -29,19 +29,14 @@
 #include <tesseract_common/macros.h>
 TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <string>
-#include <typeindex>
 #include <boost/serialization/base_object.hpp>
 #include <boost/serialization/export.hpp>
-#include <boost/serialization/nvp.hpp>
-#include <boost/serialization/unique_ptr.hpp>
-#include <boost/serialization/export.hpp>
-#include <boost/type_traits/is_virtual_base_of.hpp>
+#include <boost/concept_check.hpp>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
-#include <tesseract_common/serialization.h>
 #include <tesseract_command_language/core/waypoint.h>
 #include <tesseract_common/serialization.h>
-#include <tesseract_common/sfinae_utils.h>
+#include <tesseract_common/type_erasure.h>
 
 #ifdef SWIG
 %ignore std::vector<tesseract_planning::Instruction>::vector(size_type);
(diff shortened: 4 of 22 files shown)


mpowelson:
This is pretty similar to the various internet questions/debates about abstract base class naming conventions. I'd probably lean suffix so that they are listed next to each other in alphabetical order and autocompletes will prompt you to select which of the 2 headers you actually want for example. However, if you find that some prominent style guide suggests otherwise, then I'm not going to be upset.

marip8:
Why do we need to know if a class is a type eraser or not? It seems like names like `MoveInstruction`, `CartesianWaypoint`, etc. are sufficiently descriptive

Levi-Armstrong:
> Why do we need to know if a class is a type eraser or not? It seems like names like `MoveInstruction`, `CartesianWaypoint`, etc. are sufficiently descriptive

The main reason is that a type erasure itself is not an implementation but an interface. If I give the type erasure class the Name `MoveInstruction` or `CartesianWaypoint` then the default implementation class must have a different name. This is similar to base class when using inheritance.

Levi-Armstrong:
Benchmark data prior to new interface
``` bash
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_InstructionPolyCreation            31.6 ns         31.6 ns     22261073
BM_WaypointPolyCreation               8.23 ns         8.23 ns     84450089
BM_CompositeInstructionCreation       77.1 ns         77.1 ns      8936164
BM_ProgramCreation                   29284 ns        29282 ns        24073
BM_InstructionPolyCopy                72.7 ns         72.7 ns      9539540
BM_WaypointPolyCopy                   21.5 ns         21.5 ns     32290055
BM_CompositeInstructionCopy           5524 ns         5523 ns       130011
BM_InstructionPolyAssign              72.4 ns         72.4 ns      9547289
BM_WaypointPolyAssign                 22.2 ns         22.2 ns     32492661
BM_CompositeInstructionAssign         5612 ns         5612 ns       124855
BM_InstructionPolyCast                2.40 ns         2.40 ns    290674813
BM_WaypointPolyCast                   2.62 ns         2.62 ns    267082974
BM_InstructionPolyAccess              2.18 ns         2.18 ns    318938027
BM_WaypointPolyAccess                 1.53 ns         1.53 ns    455424083
```

Levi-Armstrong:
Benchmark data using new interface for Instruction and Waypoint
``` bash
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_InstructionPolyCreation            29.3 ns         29.3 ns     23899770
BM_WaypointPolyCreation               7.86 ns         7.86 ns     87595788
BM_CompositeInstructionCreation       75.2 ns         75.2 ns      9170892
BM_ProgramCreation                   35205 ns        35202 ns        20198
BM_InstructionPolyCopy                63.9 ns         63.9 ns     10862966
BM_WaypointPolyCopy                   18.8 ns         18.8 ns     37556489
BM_CompositeInstructionCopy           4777 ns         4777 ns       138690
BM_InstructionPolyAssign              64.4 ns         64.4 ns     10773868
BM_WaypointPolyAssign                 18.6 ns         18.6 ns     37553212
BM_CompositeInstructionAssign         4878 ns         4877 ns       135252
BM_InstructionPolyCast                1.97 ns         1.97 ns    355673505
BM_WaypointPolyCast                   1.96 ns         1.96 ns    355315993
BM_InstructionPolyAccess              2.18 ns         2.18 ns    320402241
BM_WaypointPolyAccess                0.327 ns        0.327 ns   1000000000
```

Levi-Armstrong:
Interesting the only difference between the two related to construction is the constructor. I switched back to the original constructor for the type erasure which appears to be the reason for the slow down on the creation of the program.

Update Benchmark data, @mpowelson good call on suggesting defining benchmarks!
``` bash
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_InstructionPolyCreation            29.1 ns         29.1 ns     24348718
BM_WaypointPolyCreation               8.51 ns         8.51 ns     86880104
BM_CompositeInstructionCreation       73.7 ns         73.7 ns      9355323
BM_ProgramCreation                   26281 ns        26279 ns        26855
BM_InstructionPolyCopy                66.0 ns         66.0 ns     10425975
BM_WaypointPolyCopy                   19.2 ns         19.2 ns     36110896
BM_CompositeInstructionCopy           4801 ns         4800 ns       150118
BM_InstructionPolyAssign              65.2 ns         65.2 ns     10530107
BM_WaypointPolyAssign                 18.8 ns         18.8 ns     37231487
BM_CompositeInstructionAssign         4792 ns         4792 ns       150193
BM_InstructionPolyCast                1.75 ns         1.75 ns    397629002
BM_WaypointPolyCast                   1.75 ns         1.75 ns    400767080
BM_InstructionPolyAccess              2.18 ns         2.18 ns    319270389
BM_WaypointPolyAccess                0.352 ns        0.352 ns   1000000000
```

Levi-Armstrong:
The original constructor
``` c++
  template <typename T, generic_ctor_enabler<T> = 0>
  TypeErasureBase(T&& value)  // NOLINT
    : value_(std::make_unique<TypeErasureInstanceWrapper<Instance<uncvref_t<T>>>>(value))
  {
  }
```
Format causing the slow down
``` c++
  template <typename T, generic_ctor_enabler<T> = 0>
  TypeErasureBase(T value)  // NOLINT
    : value_(std::make_unique<TypeErasureInstanceWrapper<Instance<T>>>(std::move(value)))
  {
  }
```

Levi-Armstrong:
I also added a few benchmarks comparing a vector of type erasures and vector of unique_pointers. A good thing is it does not look like there is a noticeable performance difference between two. 

``` bash
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
BM_InstructionPolyCreation               30.9 ns         30.9 ns     22914985
BM_WaypointPolyCreation                  8.23 ns         8.23 ns     85001327
BM_CompositeInstructionCreation          81.2 ns         81.2 ns      8612000
BM_ProgramCreation                      28991 ns        28990 ns        24260
BM_InstructionPolyCopy                   79.5 ns         79.5 ns      8685787
BM_WaypointPolyCopy                      19.2 ns         19.2 ns     36459752
BM_CompositeInstructionCopy              5384 ns         5385 ns       129476
BM_InstructionPolyAssign                 79.8 ns         79.8 ns      8705619
BM_WaypointPolyAssign                    19.1 ns         19.1 ns     36217834
BM_CompositeInstructionAssign            5443 ns         5443 ns       129503
BM_InstructionPolyCast                   1.76 ns         1.76 ns    398397808
BM_WaypointPolyCast                      1.76 ns         1.76 ns    394882679
BM_InstructionPolyAccess                 2.19 ns         2.19 ns    319837789
BM_WaypointPolyAccess                   0.358 ns        0.358 ns   1000000000
BM_VectorStateWaypointPolyCreation     143354 ns       143352 ns         4941
BM_VectorStateWaypointUPtrCreation     143698 ns       143698 ns         4891
BM_VectorStateWaypointPolyCopy          88228 ns        88228 ns         7596
BM_VectorStateWaypointUPtrCopy          87219 ns        87219 ns         7952
```

marip8:
> The main reason is that a type erasure itself is not an implementation but an interface. If I give the type erasure class the Name MoveInstruction or CartesianWaypoint then the default implementation class must have a different name. This is similar to base class when using inheritance.

I'm not sure I fully understand. Base classes and interfaces don't have to be pure virtual; they can be fully implemented. It seems like the Cartesian waypoint for example is both a fully implemented waypoint type and is capable of being inherited from. Why do you need to change its name? It seems like that would just introduce confusion. Cartesian waypoint itself is also still just an implementation of the `Waypoint` interface too right?

mpowelson:
>Interesting the only difference between the two related to construction is the constructor. I switched back to the original constructor for the type erasure which appears to be the reason for the slow down on the creation of the program.

Nice! I realize these were run at different times so it could just be that your computer was running faster the second time, but it looks like the new ones are generally faster for assignment and casting too.

Levi-Armstrong:
> I'm not sure I fully understand. Base classes and interfaces don't have to be pure virtual; they can be fully implemented. It seems like the Cartesian waypoint for example is both a fully implemented waypoint type and is capable of being inherited from. Why do you need to change its name? It seems like that would just introduce confusion. Cartesian waypoint itself is also still just an implementation of the `Waypoint` interface too right?

In case of the type erasure definition you cannot have a default implementation. Now you can have it default construction with another type like what we are currently doing with the null instruction or null waypoint, but this must be another class. In order for us to enable the ability to support user defined MoveInstruction, CartesianWaypoint, JointWaypoint and StateWaypoint these must all become type erasure definitions so the user can define there own as long as it matches the right interface and the planners with operate using the type erasure type.

Levi-Armstrong:
Based on discussion offline we are going to suffix with Poly. I will follow up with a PR that switched the necessary types to Type Erasures.

PR #207, opened 2022-06-29
Add support for sco::Optimizer::Callbacks to the trajopt solver profile

Levi-Armstrong:


cpetersmeier:
Thanks a lot! Can we just omit the callbacks in the `clear()` function as they are not part of the motion planner class anymore like so: https://github.com/cpetersmeier/tesseract_planning/commit/d8d1d22da427527ba30dd4b2636e422491586936 ? I think that should make the tests pass.

PR #208, opened 2022-06-29
CPack

marip8:
This PR adds CPack to the packages in this repository. I've tested with Ubuntu Focal but not yet with Windows

### DIFF
--- a/tesseract_motion_planners/CMakeLists.txt
+++ b/tesseract_motion_planners/CMakeLists.txt
@@ -87,3 +87,69 @@ endif()
 
 # Package configuration
 configure_package(NAMESPACE tesseract)
+
+if(TESSERACT_PACKAGE)
+  set(LINUX_DEPENDS
+      "libconsole-bridge-dev"
+      "libeigen3-dev"
+      "libboost-dev"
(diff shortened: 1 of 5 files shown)


tesseract_motion_planners/CMakeLists.txt
```diff
@@ -87,3 +87,69 @@ endif()
...
+      APPEND
+      WINDOW_DEPENDS
+      "descartes_light"
+      "${TESSERACT_PACKAGE_PREFIX}tesseract-collision")
+  endif()
+  if(TESSERACT_BUILD_TRAJOPT)
+    list(
+      APPEND
+      LINUX_DEPENDS
+      "trajopt"
```
marip8, line 126: Do you think we should prefix these with `${TRAJOPT_PACKAGE_PREFIX}` in the case someone builds these packages with a non-empty prefix (e.g. `ros-noetic-`)?

Levi-Armstrong: This way you could have both installed if you wanted.

Levi-Armstrong (approved):

PR #209, opened 2022-06-29
Set the default descartes behavior to treat all states equally

marrts:
Previously the default state evaluator behavior would apply a weight to every state based on the sum of the squares of the joint values. This is okay in some instances, but should not be the default behavior. This can lead to an "optimal" Descartes solution that favors a particular joint configuration for longer than it should through a trajectory and then a sharp change at the end only when necessary. Also, if a system has a rail or gantry it favors values of 0, which shouldn't be default behavior.

Depends on https://github.com/swri-robotics/descartes_light/pull/94.


marrts:
The required changes have been merged into Descartes, but the dependencies are pointing towards, Descartes version 0.3.0, so this will fail CI. @Levi-Armstrong, not sure what course of action you want me to take.

marip8:
> Would you like me to create a new release of descartes?

Sure. I was hoping to get the updated CI builds merged in first, but we can push another tag once we get that figured out. Is there anything special about the descartes release process other than pushing a tag?

Levi-Armstrong:
@marrts If you update the descartes tag it should pass CI

PR #212, opened 2022-06-30
Renames in type erasure to avoid WIN32 defines

johnwason:
This updates tesseract_planning to https://github.com/tesseract-robotics/tesseract/pull/790


johnwason:
These tests should be rerun

Levi-Armstrong:
@johnwason It looks like the PR against tesseract does not incldue the `getInterface` name change.

johnwason:
I see the change in master: https://github.com/tesseract-robotics/tesseract/blob/ce2058a874446a7cdc536bdd8b180d85526257f9/tesseract_common/include/tesseract_common/type_erasure.h#L197

PR #213, opened 2022-07-04
Add ability to submit ProcessPlanningProblem to planning server

Levi-Armstrong:


tesseract_process_managers/src/core/process_planning_server.cpp
```diff
@@ -205,27 +224,28 @@ ProcessPlanningFuture ProcessPlanningServer::run(const ProcessPlanningRequest& r
...
+                  tesseract_common::getTimestampString() + ".dot");
     response.problem->taskflow_container.taskflow->dump(out_data);
     out_data.close();
   }
 
   std::shared_ptr<tf::Executor> executor;
   {
     std::shared_lock lock(mutex_);
-    executor = executors_.at(request.executor_name);
+    executor = executors_.at(name);
```
mpowelson, line 244: Should we be checking that this exists?

tesseract_process_managers/src/core/process_planning_server.cpp
```diff
@@ -144,56 +144,75 @@ std::vector<std::string> ProcessPlanningServer::getAvailableProcessPlanners() co
...
 {
-  CONSOLE_BRIDGE_logInform("Tesseract Planning Server Received Request!");
-  ProcessPlanningFuture response;
-  response.problem->name = request.name;
-  response.problem->plan_profile_remapping =
-      std::make_unique<const PlannerProfileRemapping>(request.plan_profile_remapping);
-  response.problem->composite_profile_remapping =
+  CONSOLE_BRIDGE_logDebug("Tesseract Planning Server Received Request!");
+
+  auto it = process_planners_.find(request.name);
```
mpowelson, line 149: Does this need to be here? This check is also in `ProcessPlanningServer::run(ProcessPlanningProblem::Ptr problem, ...` so it should get checked before execution. This will check it twice.

tesseract_process_managers/src/core/process_planning_server.cpp
```diff
@@ -144,56 +144,75 @@ std::vector<std::string> ProcessPlanningServer::getAvailableProcessPlanners() co
...
+  auto problem = std::make_shared<ProcessPlanningProblem>();
+  problem->name = request.name;
+  problem->plan_profile_remapping = std::make_unique<const PlannerProfileRemapping>(request.plan_profile_remapping);
+  problem->composite_profile_remapping =
       std::make_unique<const PlannerProfileRemapping>(request.composite_profile_remapping);
 
-  response.problem->input = std::make_unique<Instruction>(request.instructions);
-  auto& composite_program = response.problem->input->as<CompositeInstruction>();
+  problem->input = std::make_unique<Instruction>(request.instructions);
+  auto& composite_program = problem->input->as<CompositeInstruction>();
```
mpowelson, line 163: `const auto&`?

tesseract_process_managers/src/core/process_planning_server.cpp
Levi-Armstrong: Added.

tesseract_process_managers/src/core/process_planning_server.cpp
Levi-Armstrong: Yea, probably can be removed.

PR #221, opened 2022-07-12
motion_planners: support util function templates

cpetersmeier:
Util functions in motion planners package to support the new templating interface.


Levi-Armstrong:
This is captured in PR #219

PR #227, opened 2022-07-24
Move most SWIG commands to tesseract_python package

johnwason:
Like https://github.com/tesseract-robotics/tesseract/pull/809 this remove SWIG command so they can be moved to the tesseract_python package.


johnwason:
The CI errors are due to isNull not being available in the tagged version of tesseract.

Levi-Armstrong:
Yea, as long as unstable passes I will merge.

PR #236, opened 2022-08-08
Add tesseract_task_composer package to replace tesseract_process_managers

Levi-Armstrong:
Highlights:
- Creates a centralized location for task data storage and retrieval
 - Abstracts away taskflow to allow implementing different libraries
 - Simplify the program structure where the output is a modified input program instead of having to maintain two different data structures
 - Move instructions have uuid  and parent uuid to allow tracking intructions and if they were derived from another instructions. For example if the simple planner interpolates the interpolated move instruction with have a parent uuid.
 - Removed the use tesseract_common::StatusCode.
 - Everything in tesseract_task_composer has boost serialization.

Note:
   - With the new data structure it simplifies the motion planners
   - For the purpose of testing I have kept the original objects associated with tesseract_process_managers where have been renamed to include legacy. This is to support testing while changes are being made. After this PR is ready to be merged a commit will be added to remove them.

Next Steps:

- [x] Create an example implementation of a raster task flow.
- [x] Finish the TaskComposerExecutor implementation
  - [x] Create default implementation using taskflow 
- [x] Port tesseract_process_managers unit tests
- [x] Remove clone method from TaskComposerNode (It was not needed and complicates things)
- [x] Finish TaskComposerServer

@marip8, @marrts, @mpowelson I recommend taking a look at the example to get an understanding how it works. In a lot of ways it is very similar to tesseract_process_managers with a lot of things simplified.


Levi-Armstrong:
@marip8 So I split up the task to avoid having configuration parameters. Take a look and let me know what you think. I will be looking storing the results in the task info and retrieve them from the task info container next.

PR #240, opened 2022-08-15
Add Eigen alignment macro to classes that contain SceneState

mpowelson:
SceneState is an aligned class, so these classes that contain a SceneState should also be aligned.


tesseract_motion_planners/core/include/tesseract_motion_planners/core/types.h
```diff
@@ -89,6 +93,10 @@ struct PlannerRequest
 
 struct PlannerResponse
 {
+  // LCOV_EXCL_START
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+  // LCOV_EXCL_STOP
+
```
mpowelson, line 99: This one is actually because CompositeInstruction is aligned

Levi-Armstrong (approved):

PR #255, opened 2022-10-11
Profile Dictionary Update

marip8:
## Changes
- Replaced profile overrides with per-instruction remapping of profile names by namespace
  - The instruction `getProfile` method now accepts a namespace argument that defaults to an empty string. If the string is empty, the nominal profile is returned. If the namespace string is not empty, the overridden profile name is returned if it exists; otherwise the nominal profile name is returned
- Consolidated profile dictionary getter utilities with single `getProfile` function:
  - `getProfile` no longer returns a default profile. An exception is thrown if the named profile of the specified type cannot be found. 
  - Error reporting was improved to make it easier to understand why the profile wasn't found
- Updated the motion planners to access profiles using the updated API
- Updated the task composer nodes to access profiles using the updated API
- Removed profile remapping from task composer input
- Updated unit tests
- Updated examples

## Background
The move and composite instructions currently have a dependency on the profile dictionary which causes issues for the upcoming introduction of new profile interfaces and an updated profile dictionary (because the interfaces require information in move and composite instructions to produce motion planning profiles). Also, conceptually the profile dictionary belongs in the motion planning module. Most of this PR addresses the replacement of the current profile overrides (which were specified as a separate profile dictionary with different keys) with an optional profile name remapping for individual instructions.

After revising the profile dictionary, I also tried to simplify the API by consolidating several external helper functions (`getProfileString`, `applyProfileOverrides`, etc.) and add more error information for when profiles can't be found. I also removed the behavior of the profile dictionary of returning a default profile when the queried profile cannot be found. This is a change in the nominal behavior, but will help significantly in understanding whether the specified profiles were actually found and used during motion planning. In updating the examples, I found that most of the profile names specified in the program instructions did not have corresponding objects in the profile dictionary and that an arbitrary default profile was being used instead. I also added two helper functions for adding default planner and task composer profiles to a dictionary for the case where the default profiles are sufficient.


marip8:
@Levi-Armstrong this PR should be ready for review. 

The `tesseract_examples/basic_cartesian_example` unit tests are failing. On my machine, I get the following error when it constructs the octomap from the point cloud:

```
tesseract_examples_basic_cartesian_example: /build/octomap-KdBKXz/octomap-1.9.3+dfsg/octomap/include/octomap/OcTreeDataNode.hxx:69: octomap::OcTreeDataNode<T>::~OcTreeDataNode() [with T = float]: Assertion `children == NULL' failed.
```
Any ideas on what might cause this? It doesn't seem to be related to any of the changes I made

Levi-Armstrong:
Here is a [PR](https://github.com/OctoMap/octomap/pull/288/files#diff-45151bec6492789e0b7d1128eb0a8488442d65ccc2756d631d68219c2a8cd180) which fixes the issue. I think we need to add the same logic to the pruneNode in tesseract_geometry

marip8:
> I tried adding the logic to our code but the object is protected so that is not going to work. Did any headers get arranged?

What object? The profile override setters and getters in the composite and move instructions should be public. I don't think I changed any headers

Levi-Armstrong:
I tried to add the following in the prune function, but children is not accessible. 
```
        if (node->children != NULL){
          delete[] node->children;
          node->children = NULL;
        }
```

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
```diff
@@ -191,11 +159,50 @@ class ProfileDictionary
   std::shared_ptr<const ProfileType> getProfile(const std::string& ns, const std::string& profile_name) const
   {
     std::shared_lock lock(mutex_);
-    const auto& it = profiles_.at(ns);
-    const auto& it2 = it.at(std::type_index(typeid(ProfileType)));
+
```
Levi-Armstrong, line 162: I am not sure how I feel about try,catch in the motion planning pipeline because it a huge time penalty when thrown an caught. Is there a reason to catch an rethrow and not let the exception flow up?

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
```diff
@@ -156,31 +149,6 @@ class ProfileDictionary
     }
   }
 
-  /**
-   * @brief Check if a profile exists
-   * @details If profile entry does not exist it also returns false
-   * @return True if profile exists, otherwise false
-   */
-  template <typename ProfileType>
-  bool hasProfile(const std::string& ns, const std::string& profile_name) const
```
Levi-Armstrong, line 165: How do you know check if a profile exists?

tesseract_motion_planners/examples/raster_example.cpp
```diff
@@ -112,14 +113,14 @@ int main(int /*argc*/, char** /*argv*/)
                                                   Eigen::Quaterniond(0, 0, -1.0, 0));
 
     // Define raster move instruction
-    MoveInstruction plan_c0(wp2, MoveInstructionType::LINEAR, "RASTER");
```
Levi-Armstrong, line 115: It looks like all of he profiles were removed. Was there a reason behind this?

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
Levi-Armstrong: I think this should remain. Using exceptions to check if something exists is very expensive.

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
Levi-Armstrong: I think we should let the this flow up and not catch an rethrow.

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -358,13 +358,7 @@ OMPLProblemConfig OMPLMotionPlanner::createSubProblem(const PlannerRequest& requ
...
 
   // Get Plan Profile
-  std::string profile = end_instruction.getProfile();
-  profile = getProfileString(name_, profile, request.plan_profile_remapping);
-  auto cur_plan_profile =
-      getProfile<OMPLPlanProfile>(name_, profile, *request.profiles, std::make_shared<OMPLDefaultPlanProfile>());
-  cur_plan_profile = applyProfileOverrides(name_, profile, cur_plan_profile, end_instruction.getProfileOverrides());
-  if (!cur_plan_profile)
-    throw std::runtime_error("OMPLMotionPlanner: Invalid profile");
+  auto cur_plan_profile = request.profiles->getProfile<OMPLPlanProfile>(name_, end_instruction.getProfile(name_));
```
Levi-Armstrong, line 361: Do you have an explanation of how this works compared to the previous implementation?

tesseract_motion_planners/test/profile_dictionary_tests.cpp
```diff
@@ -58,30 +60,28 @@ TEST(TesseractPlanningProfileDictionaryUnit, ProfileDictionaryTest)  // NOLINT
 {
   ProfileDictionary profiles;
 
-  EXPECT_FALSE(profiles.hasProfileEntry<ProfileBase>("ns"));
+  EXPECT_THROW(profiles.getProfile<ProfileBase>("ns", "key"), std::runtime_error);  // NOLINT
```
Levi-Armstrong, line 63: I think we should switch back to using the hasProfile.

Levi-Armstrong (changes requested):
I am about half way through the review. I think we should not hold this up due to the unit test failure. I am going to look into it this weekend and let you know what I find.

Levi-Armstrong:
@marip8 Do you have a tesseract_ros branch with updates?

Levi-Armstrong:
@marip8 I was getting a different issue when running the example where it was failing due to missing profiles. After updating the cartesian example everything passes locally. I pushed my commit to your branch so we'll see if the tests pass now.

Levi-Armstrong:
It looks like CI is passing now with my latest commit. I will finish my review early this week.

tesseract_task_composer/src/nodes/time_optimal_parameterization_task.cpp
```diff
@@ -113,20 +110,18 @@ TaskComposerNodeInfo::UPtr TimeOptimalParameterizationTask::runImpl(TaskComposer
...
-
-    // Check for remapping of the plan profile
-    std::string remap = getProfileString(name_, profile, input.problem.move_profile_remapping);
-    auto cur_move_profile = getProfile<TimeOptimalParameterizationProfile>(
-        name_, remap, *input.profiles, std::make_shared<TimeOptimalParameterizationProfile>());
-    cur_move_profile = applyProfileOverrides(name_, remap, cur_move_profile, mi.getProfileOverrides());
 
     // If there is a move profile associated with it, override the parameters
-    if (cur_move_profile)
+    try
```
Levi-Armstrong, line 115: same here

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
```diff
@@ -191,11 +159,50 @@ class ProfileDictionary
   std::shared_ptr<const ProfileType> getProfile(const std::string& ns, const std::string& profile_name) const
   {
     std::shared_lock lock(mutex_);
-    const auto& it = profiles_.at(ns);
-    const auto& it2 = it.at(std::type_index(typeid(ProfileType)));
+
+    const std::unordered_map<std::type_index, std::any>* it = nullptr;
+    try
```
Levi-Armstrong, line 164: I would avoid a catch here and just check if it exists and if not throw exception.

Levi-Armstrong (changes requested):

tesseract_motion_planners/examples/raster_example.cpp
marip8: None of them were ever actually added to the profile dictionary, so it was just pulling the default profile. That was the case in most of the examples actually. A lot of the changes I made in this PR were related to raising exceptions when requested profile names didn't have a corresponding profile entry in the dictionary.

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
marip8: The same. `end_instruction.getProfile(name_)` attempts to get the profile name associated with the specific planner (i.e. `name_`); if that doesn't exist, the default profile name is returned. Then `getProfile` tries to find a profile of the specified type with the profile name returned. If it does not exist, it throws a `std::runtime_error`

marip8:
It seems like most of your comments are around throwing exceptions. These changes were mostly patches to the existing profile dictionary to remove the behavior of returning of default profiles when no profiles were actually added to the dictionary for a specific profile name. When we change the profile interfaces, we will introduce a new profile dictionary and the changes to the existing profile dictionary will go away. I would prefer not to reintroduce `hasProfile` here, because it seems like wasted effort. We can keep exception throwing in mind in the next profile dictionary.

I personally don't see the exceptions in this PR as a huge performance hit because most of them are only caught in order to add more information and then be rethrown. The only place where that is not the case is in the time parameterization tasks, and we could change that in future with a `hasProfile` method.

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
marip8: Both catches here simply rethrow the exception with more, relevant information. We could do a check with `it->find(...)` and then throw an exception if the key does not exist, but that's functionally the same as what is here already. In either case, a terminating exception gets thrown and (almost) nothing downstream in the planning framework catches it, so it doesn't seem like the catch is really that expensive

marip8:
> @marip8 Do you have a tesseract_ros branch with updates?

No I do not

Levi-Armstrong:
> It seems like most of your comments are around throwing exceptions. These changes were mostly patches to the existing profile dictionary to remove the behavior of returning of default profiles when no profiles were actually added to the dictionary for a specific profile name. When we change the profile interfaces, we will introduce a new profile dictionary and the changes to the existing profile dictionary will go away. I would prefer not to reintroduce `hasProfile` here, because it seems like wasted effort. We can keep exception throwing in mind in the next profile dictionary.
> 
> I personally don't see the exceptions in this PR as a huge performance hit because most of them are only caught in order to add more information and then be rethrown. The only place where that is not the case is in the time parameterization tasks, and we could change that in future with a `hasProfile` method.

Sounds good.

marip8:
Are there any planning time benchmarks? If not, it would be interesting to try to convert some of the planning examples to benchmarks where we can measure planning time. Then it could help us justify decisions about throwing or not throwing exceptions. My hunch is that the exceptions will generally only be thrown in cases where a planning problem is not set up correctly and they won't otherwise be thrown, so we won't see a time penalty

tesseract_motion_planners/trajopt_ifopt/src/trajopt_ifopt_motion_planner.cpp
```diff
@@ -313,7 +301,10 @@ std::shared_ptr<TrajOptIfoptProblem> TrajOptIfoptMotionPlanner::createProblem(co
           fixed_steps.push_back(i);
       }
     }
-    else if (move_instruction.getWaypoint().isStateWaypoint())
+    else if (move_instruction.getWaypoint().isStateWaypoint())  //  solver_profile = applyProfileOverrides(name,
```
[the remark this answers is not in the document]
marip8: Feel free to remove yourself today if you're working on this. I won't be able to get to it until tomorrow morning or possibly Wednesday

Levi-Armstrong:
> My hunch is that the exceptions will generally only be thrown in cases where a planning problem is not set up correctly and they won't otherwise be thrown, so we won't see a time penalty

I don't think we currently have any in tesseract_planning. I have done internal benchmarking with exception and to use them for logical flow like what is done in time parameterization it is significant. It is only significant when an exception is thrown an caught. If no exception is thrown then you will not have a penalty so as long as they are not being used for logical flow they should be fine.

Levi-Armstrong:
I found my messages. In my test case using godbolt it was 100 times slower to use exceptions for logical flow versus `if, else` for a simple test case. I would assume this increase based on the stack size it has to unwind. 

```
18.2991 if, else
18.1554 try, catch no exception thrown
2456.4 try, catch exception thrown
```

https://mortoray.com/the-true-cost-of-zero-cost-exceptions/

tesseract_task_composer/src/nodes/ruckig_trajectory_smoothing_task.cpp
```diff
@@ -114,21 +111,19 @@ TaskComposerNodeInfo::UPtr RuckigTrajectorySmoothingTask::runImpl(TaskComposerIn
...
-
-    // Check for remapping of the plan profile
-    move_profile = getProfileString(name_, profile, input.problem.move_profile_remapping);
-    auto cur_move_profile = getProfile<RuckigTrajectorySmoothingMoveProfile>(
-        name_, move_profile, *input.profiles, std::make_shared<RuckigTrajectorySmoothingMoveProfile>());
-    //    cur_move_profile = applyProfileOverrides(name_, profile, cur_move_profile, mi.profile_overrides);
 
     // If there is a move profile associated with it, override the parameters
-    if (cur_move_profile)
+    try
```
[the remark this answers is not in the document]
Levi-Armstrong: This will be addressed in profile refactor

tesseract_task_composer/src/nodes/time_optimal_parameterization_task.cpp
Levi-Armstrong: Note: This will be addressed in profile refactor

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
Levi-Armstrong: Yea, it seem this is something that should not happen in normal operation.

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
Levi-Armstrong: Note: This will be addressed in profile refactor

tesseract_command_language/include/tesseract_command_language/profile_dictionary.h
Levi-Armstrong: Note: This will be addressed in profile refactor

tesseract_motion_planners/test/profile_dictionary_tests.cpp
Levi-Armstrong: Note: This will be addressed in profile refactor

tesseract_motion_planners/trajopt_ifopt/src/trajopt_ifopt_motion_planner.cpp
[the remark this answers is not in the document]
Levi-Armstrong: fixed

Levi-Armstrong (approved):

Levi-Armstrong:
@marip8 Do you want me to squash merge this?

marip8:
That works for me; this should be complete as far as this PR goes

Levi-Armstrong:
@marip8 It is merged into feat/ProfileRefactor which you can use to create PR's against as you work on the Profile Refactoring.

PR #267, opened 2023-01-12
Revised planner profile interfaces

marip8:
This PR provides new interfaces for motion planner waypoint, composite, and planner level profiles. Each profile produces a type-erased, planner-specific data object from a relevant input (an instruction for the waypoint profile, a composite instruction for the composite profile). The individual planners can look up profiles in the profile dictionary by a combination of namespace and name, and can cast them to the required type. Since the objects returned by the profiles are std::any type-erasers, any issues with incorrect casts will result in an exception that can be handled.

The addition of new planner profile interfaces supports an on-going larger refactor and simplification of the various motion planners class (started in https://github.com/tesseract-robotics/tesseract_planning/pull/138). Additionally, the new profiles will be easier to store and retrieve in the profile dictionary since they will no longer be stored by class typeid.

This PR adds three new maps in the profile dictionary to hold these waypoints. Currently these maps are public members of the profile dictionary for convenience and to minimize API breakage. As motion planners are refactored to use these profile interfaces, the planners will retrieve profiles from these maps. Ultimately, the map storing profiles by typeid will be removed, the three maps holding the new profiles will be made private, and the existing profile dictionary API will be revised to access the new profile maps.

Replaces #156

### DIFF
--- a/tesseract_command_language/test/move_instruction_unit.cpp
+++ b/tesseract_command_language/test/move_instruction_unit.cpp
@@ -296,7 +296,7 @@ TEST(TesseractCommandLanguageMoveInstructionPolyUnit, ProfileOverrides)  // NOLI
   Eigen::VectorXd jv = Eigen::VectorXd::Ones(6);
   std::vector<std::string> jn = { "j1", "j2", "j3", "j4", "j5", "j6" };
   StateWaypointPoly swp(StateWaypoint(jn, jv));
-  MoveInstruction instr(swp, MoveInstructionType::START, DEFAULT_PROFILE_KEY, DEFAULT_PROFILE_KEY);
+  MoveInstruction instr(swp, MoveInstructionType::FREESPACE, DEFAULT_PROFILE_KEY, DEFAULT_PROFILE_KEY);
 
   // Create arbitrary profile overrides under arbitrary namespaces
   const std::string ns1 = "ns1";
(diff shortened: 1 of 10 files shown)


Levi-Armstrong:
LGTM, feel free to merge when you are ready.

marip8:
What did you have to do on the master branch to prevent the Bionic build from failing? It complains about not knowing about a `scoped_lock`, but the builds aren't failing on the master branch. I also noticed that the last PR we merged into this branch also did not have a successful Bionic build for the same reason

Levi-Armstrong:
Maybe we just need to rebase this on master

Levi-Armstrong:
I rebased feat/ProfileRefactor on master. Rebase yours on feat/ProfileRefactor again and lets see if the CI issue goes away. If not after that then I would say just let it go for now and we track down the issue later since Focal is passing.

marip8:
I think the rebase of #255 and `feat/ProfileRefactor` onto the current master messed up some of the unit tests since you also changed them in the recently merged PRs. These should be fixed now

We should also seriously consider merging #255 and the last two commits here ([a40f272](https://github.com/tesseract-robotics/tesseract_planning/pull/267/commits/a40f272c2c7aa7da0b0d85b3c6145bb694588fbc) and [bcc8f8e](https://github.com/tesseract-robotics/tesseract_planning/pull/267/commits/a40f272c2c7aa7da0b0d85b3c6145bb694588fbc)) into the master branch. These examples were failing because profiles for the specified profile names of "FREESPACE" and "UPRIGHT" were never actually added to the profile dictionary, and it was silently using the default profiles.

PR #279, opened 2023-02-10
Use templates for raster task to reduce code duplications

Levi-Armstrong:
@marrts This is a start to what I was talking about with leveraging templates.


marrts:
This makes a lot of sense. So you have base, templated, implementations of each type of planner and then a few defaults (ct and ft) that have preset types.

As a side note, the default global descartes ones don't really work as intended because the default cartesian planner runs descartes. So the pipeline runs a global descartes, followed by a descartes on each individual raster which overrides the joint solution from the global plan. You need to use a cartesian pipeline that just runs trajopt.

Levi-Armstrong:
>This makes a lot of sense. So you have base, templated, implementations of each type of planner and then a few defaults (ct and ft) that have preset types.

Yea, now you do not need to create a separate class if you do not need serialization you can just instantiate using template class directly. 

> As a side note, the default global descartes ones don't really work as intended because the default cartesian planner runs descartes. So the pipeline runs a global descartes, followed by a descartes on each individual raster which overrides the joint solution from the global plan. You need to use a cartesian pipeline that just runs trajopt.

Good catch. I will updated the global variants so they do not run Descartes again.

tesseract_task_composer/include/tesseract_task_composer/nodes/raster_motion_task.hpp
```diff
@@ -0,0 +1,336 @@
...
+ */
+template <typename FreespaceTaskType, typename RasterTaskType, typename TransitionTaskType>
+class RasterMotionTask : public TaskComposerTask
+{
+public:
+  using class_type = RasterMotionTask<FreespaceTaskType, RasterTaskType, TransitionTaskType>;
+
+  RasterMotionTask() = default;  // Required for serialization
+  // NOLINTNEXTLINE(performance-unnecessary-value-param)
+  RasterMotionTask(std::string input_key, std::string output_key, bool is_conditional, std::string name)
```
marrts, line 72: You removed the defaults for is_conditional and name. Later when calling it you only set the is_conditional parameter. You need to fix this at one place and/or the other. See: https://github.com/tesseract-robotics/tesseract_planning/blob/c32e377106247d278f15a37a543c1361a52a32f6/tesseract_task_composer/include/tesseract_task_composer/nodes/raster_global_pipeline_task.hpp#L56

marrts:
Also there is no way to get the underlying `task_graph object from the `RasterMotionTask` type. The graph is built and run all inside of the `runImpl`, which works, but you can't ever visualize the actual graph, I believe before it was set up to always save it to tmp with a hardcoded name, which is also not a great solution.

marip8:
It seems like a better approach to composable tasks/graphs like this would be to use a plugin loader to load tasks dynamically, rather than defining template functions that need to be compiled for every combination of planner inputs. This would apply to all of the 10+ raster tasks and other free-space and Cartesian planner task flows that currently exist. I would rather spend development effort on an approach like that rather than this template-based approach that will likely be replaced by the plugin-style architecture in the long run anyway

tesseract_task_composer/include/tesseract_task_composer/nodes/raster_motion_task.hpp
Levi-Armstrong: It should just be using the default name. Did you run into an issue?

Levi-Armstrong:
> It seems like a better approach to composable tasks/graphs like this would be to use a plugin loader to load tasks dynamically, rather than defining template functions that need to be compiled for every combination of planner inputs. This would apply to all of the 10+ raster tasks and other free-space and Cartesian planner task flows that currently exist. I would rather spend development effort on an approach like that rather than this template-based approach that will likely be replaced by the plugin-style architecture in the long run anyway. It feels like the level of effort for introducing plugins to the motion planners and tasks is not very high

I agree this is where we want to end up, but it is more work than I want to take on for the next release. I think this would provide most of the past capabilities for now. Then I want to focus next on getting your profile work in and after that focus on the plugin work.

tesseract_task_composer/include/tesseract_task_composer/nodes/raster_motion_task.hpp
marrts: Sorry, just now seeing this after I got the merge notification. The default setting of the names in the base class were removed, so it fails when you try and make a custom task node implementation and don't define the name. Also the implementation of the global planner explicitly does not include a setting of the name to the templated RasterTaskType in the above segment of code I referenced, so this will always fail.

PR #281, opened 2023-02-17
Fix Default Raster Base Name

marrts:
Fixed issue relating to global planner being unusable since no raster motion task name was set


marrts:
@Levi-Armstrong Not sure exactly how you want to handle this. Either we can do some variation of assigning a default name like I do here. Or you can change the following section to pass a name string. https://github.com/tesseract-robotics/tesseract_planning/blob/2c0aa9e562298b240ec3183acb0d47a872527100/tesseract_task_composer/include/tesseract_task_composer/nodes/raster_global_pipeline_task.hpp#L56

Levi-Armstrong:
This will be addressed by PR #282

PR #282, opened 2023-02-19
Update task composer to leverage plugins

Levi-Armstrong:
@marip8 Here is a start to adding plugin support. I have the CartesianMotionPipeline implemented within the yaml config. I will be working on getting the rest moved over to being defined within the yaml config and deleting the c++ versions.


Levi-Armstrong:
@marip8 Plugin in support is implemented for executors and nodes/tasks. In addition the task_composer_server can load these from a config and I will have a PR against tesseract_ros which leverages the default config file. I did not create the secondary tasks for basic pipeline which I think we can add later if needed.

I think next steps is to get your profile work completed and I think things will be in a good place. Using the config it is easier to have issues with tasks using the default when it should not be so it is probably better to get this in soon as we can.

PR #297, opened 2023-03-25
Update to support new contact results class

Levi-Armstrong:
See tesseract [PR](https://github.com/tesseract-robotics/tesseract/pull/869) for more information.


Levi-Armstrong:
@marip8 There seems to be an issue with focal-master because it does not have the latest code because unstable is failing.

PR #298, opened 2023-03-28
This fixes crash in ompl::geometric::SimpleSetup::simplifySolution()

afrixs:
Occurs in freespace_hybrid_example

reproducible e.g. with
 - Ubuntu 22.04.02 LTS
 - `ros-humble-rclcpp 16.0.3-1jammy.20230302.174123`
 - `ros-humble-ompl 1.6.0-1jammy.20230117.223758`
 - `tesseract_ros2` from [marrts/0_15-parity](https://github.com/marrts/tesseract_ros2/tree/0_15-parity) branch

by calling
`ros2 launch tesseract_ros_examples freespace_hybrid_example.launch.xml`


Levi-Armstrong:
If you add it as shown below does the issue go away? 
``` c++
auto& p = pc.problem;
p->simple_setup->setup();
```

rjoomen:
@Levi-Armstrong, indeed, that's enough (tested with the freespace examples).

Levi-Armstrong:
Okay, something in OMPL must have changed. These test used to run succesfuly and the reason I was not calling `setup` is because the documentation for setup state that it is called in solve. I assume it should not hurt if it is called twice.

afrixs:
Moved `p->simple_setup->setup()` as proposed.

I think that `setup()` [is called](https://github.com/ompl/ompl/blob/e2994e580fcba0eb117e38bd1dc7439c4beb35ef/src/ompl/geometric/src/SimpleSetup.cpp#L125) inside `solve()` when actually planning, but apparently there are some scenarios when planning is not performed at all (at least when `p->planning_time` is 0 or very low)... Calling `setup()` twice doesn't hurt, there is a [guard condition](https://github.com/ompl/ompl/blob/e2994e580fcba0eb117e38bd1dc7439c4beb35ef/src/ompl/geometric/src/SimpleSetup.cpp#L56) to prevent double setup.

PR #300, opened 2023-03-31
Fixed documentation build job

marip8:
Changes per commit message


Levi-Armstrong:
Quality should use the master version because that is what new development is related to.

marip8:
> Quality should use the master version because that is what new development is related to.

Building against the `tesseract` master branch is unstable. I think the code quality jobs should be like the other Ubuntu and Windows jobs that target the corresponding stable version of `tesseract`, right? The reason the clang-tidy job is failing now is because there is an unused member parameter in TOTG

Levi-Armstrong:
> > Quality should use the master version because that is what new development is related to.
> 
> Building against the `tesseract` master branch is unstable. I think the code quality jobs should be like the other Ubuntu and Windows jobs that target the corresponding stable version of `tesseract`, right? The reason the clang-tidy job is failing now is because there is an unused member parameter in TOTG

All new development is relative to the master branches of all of its dependencies. Making code quality run for a fixed version is an issue because it is possible to always fail if it required the master branch. I think everything should use the master docker except focal which is to indicate which breaking changes have occurred and that the next release be either a minor or patch version update. I updated all of CI a while back to all use master because for a while every PR, all of the checks were failing rendering them useless.

Levi-Armstrong:
> > The reason the clang-tidy job is failing now is because there is an unused member parameter in TOTG

Yea, this is because CI was broken for a week or so. I have a PR which fixes this which should be merged soon.

marip8:
> All new development is relative to the master branches of all of its Tesseract dependencies.

Sure, but I am of the opinion that this is the purpose of the unstable build.

In my mind, the code quality build should be set up the same as the Ubuntu and Windows jobs to ensure that this repository is capable of being built on a specific version of `tesseract` and still meets the clang-tidy and code coverage specifications. 

We could add a second code quality job that runs on the `tesseract` master docker (or change the unstable build to use clang-tidy), but that won't tell you anything different than the job running on a tagged version of `tesseract` because it's still only checking quality associated with the code in this repo

> I think everything should use the master docker except focal which is to indicate when breaking changes have occurred and that the next release be either a minor or patch version update.

Using the `tesseract` master docker images won't indicate when this repository has breaking changes because `tesseract` does not depend on this repo. That type of build can only show when this repository is no longer compatible with the specified version of `tesseract`.

Levi-Armstrong:
The purpose of this repositories CI is to test this repos code changes. In the case of code coverage and clang-tidy there is no value in having it use the non master version because when there are breaking changes in dependencies that is all it will tell you. This is already covered by the Ubuntu/focal build. 

The purpose of codecov and clang-tidy is to tell you when proposed code changes fail these checks and in-order to do this all the time it needs to leverage the master docker. Sure we could add two codecov and clang-tidy but this adds no value because you would have 3 CI build telling the same thing that there is breaking changes in Tesseract.

You are correct that the focal CI build is to indicate an incompatibility between the repose which is currently used to indicate that the minor version needs to be incremented when making the next release. If all of these are successful on the Tesseract repositories then the patch version is incremented when making the next release. This is the only thing I use the focal CI builds for at the moment.

marip8:
Just reset quality build to build on `master` branch

Levi-Armstrong (approved):

Issue #301, opened 2023-04-05
Task Composer Configuration File Loading


marip8:
Is the following function intended to load the task composer config from a string filename or a YAML string? This does not appear to be tested in any unit tests yet.

https://github.com/tesseract-robotics/tesseract_planning/blob/e92083398c11da5a808794bed82cec392f2c4625/tesseract_task_composer/src/task_composer_plugin_factory.cpp#L89

If the intent is to load from a string filename (as in most applications I have encountered so far), then this function needs to use `YAML::LoadFile` instead of `YAML::Load`

Levi-Armstrong:
The header documentation state that it is loading it as a string. The method above states loading from file using `tesseract_common::fs::path`.

Levi-Armstrong:
Probably good to update the header to state yaml string to avoid confusion.

Levi-Armstrong:
I believe this can be closed. Reopen if not the case.

PR #304, opened 2023-04-07
Fixed task composer yaml loading to expect a filepath

marrts:
Previously the loading of config file was expecting a string that was a YAML file. It makes more sense to me to load in a filepath.


marip8:
Per #301, @Levi-Armstrong wants to use this method to use a YAML string, not a string file path. I think we should just rename the method or the argument or the docstring to indicate that the string is a YAML string

marrts:
Oops, should've seen that

PR #329, opened 2023-06-11
Fixes for Python wrappers

johnwason:
This PR includes fixes for the Python wrappers. It makes the following changes:

* Fix some library dependency issues in task composer
* Use `&&` when passing shared pointers as parameters
* Use vcpkg for build dependencies on Windows CI
* Use correct path for jsoncpp include files
* Use preprocessor `TESSERACT_TASK_COMPOSER_HAS_TRAJOPT_IFOPT` to disable ifopt when not available in task composer
* Add plugin anchors to task composer plugins

These updates may require https://github.com/tesseract-robotics/trajopt/pull/327 to build properly.


johnwason:
I fixed the failing boost serialization tests (bad /tmp path for windows), but now one of the cartesian waypoint tests is failing. I don't think my changes are causing that failure.

tesseract_task_composer/planning/src/planning_task_composer_plugin_factories.cpp
```diff
@@ -71,14 +74,18 @@ using DescartesDMotionPlannerTaskFactory = TaskComposerTaskFactory<MotionPlanner
...
+
+TESSERACT_PLUGIN_ANCHOR_IMPL(TaskComposerPlanningFactoriesAnchor);
 }  // namespace tesseract_planning
 
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 TESSERACT_ADD_TASK_COMPOSER_NODE_PLUGIN(tesseract_planning::CheckInputTaskFactory, CheckInputTaskFactory)
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 TESSERACT_ADD_TASK_COMPOSER_NODE_PLUGIN(tesseract_planning::ContinuousContactCheckTaskFactory,
-                                        ContinuousContactCheckTaskFactory)
+                                        ContinuousContactChheckTaskFactory)
```
Levi-Armstrong, line 88: An `h` was added and should probably be removed.

tesseract_task_composer/planning/src/planning_task_composer_plugin_factories.cpp
```diff
@@ -123,6 +130,8 @@ TESSERACT_ADD_TASK_COMPOSER_NODE_PLUGIN(tesseract_planning::TrajOptMotionPlanner
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 TESSERACT_ADD_TASK_COMPOSER_NODE_PLUGIN(tesseract_planning::SimpleMotionPlannerTaskFactory,
                                         SimpleMotionPlannerTaskFactory)
+#ifdef TESSERACT_TASK_COMPOSER_HAS_TRAJOPT_IFOPT
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 TESSERACT_ADD_TASK_COMPOSER_NODE_PLUGIN(tesseract_planning::TrajOptIfoptMotionPlannerTaskFactory,
                                         TrajOptIfoptMotionPlannerTaskFactory)
+#endif
```
Levi-Armstrong, line 137: Need to remove.

Levi-Armstrong (changes requested):
Just curious, why is the `&&` change required?

tesseract_task_composer/planning/src/planning_task_composer_plugin_factories.cpp
Levi-Armstrong: It is not showing here but this file has a windows end of file

johnwason:
The `&&` affects the way that SWIG generates the wrappers. Without it, it will try to copy the value. The `&&` is the correct way to format the argument so I think it is a reasonable change.

johnwason:
I have always seen the r-value reference `&&` used with std::move when ownership is being taken. Passing by value implies that the value will be copied, not ownership taken.

Levi-Armstrong:
Below is example explain there point of view. Using std::move does not actually move the object it just returns the rvalue reference which allows it to be moved if the object has a constructor which excepts the rvalue reference. If you have a function which excepts a rvalue reference there is no guarantee that the object is actually moved. In the example below it shows an example of passing it by rvalue reference but not actual moved so you can still operate on it in the main without any issues. This seem more error prone for an interface where the other does not have this issue and is guaranteed that the object is moved and cannot be operated on anymore. 


``` c++
double func1(std::unique_ptr<double>&& x)
{
    (*x)+=5;
    return (*x);
};

double func2(std::unique_ptr<double> x)
{
    (*x)+=5;
    return (*x);
};

int main()
{
    {
        auto x = std::make_unique<double>(1);
        std::cout << func1(std::move(x)) << std::endl;
        std::cout << (*x) << std::endl; // This is not an issue because the function does not take ownership
    }
    
    {
        auto x = std::make_unique<double>(1);
        std::cout << func2(std::move(x)) << std::endl;
        std::cout << (*x) << std::endl; // This is undefined behavior becaue the function does take ownership
    }
}
```

Levi-Armstrong:
Are we using SWIG 4.1 because it appears this should not be necessary if we are shown [here](https://swig.org/Doc4.1/Library.html#Library_std_unique_ptr)?

johnwason:
The built-in unique_ptr Swig support didn't work, but I was able to modify the templates to support the unique_ptr pass-by-value without needing to modify the upstream functions. This PR should be ready to merge.

Levi-Armstrong:
I will take a look tomorrow and get it merged.

Levi-Armstrong:
@johnwason I rebased on master and a new unit test is failing on windows. I will merge, but can you take a look when you have time?

Levi-Armstrong:
I will run memcheck on the unit test, but if that does not identify any issue may need help on the windows side debugging this.

johnwason:
@Levi-Armstrong it looks like pointers stored as unique_ptr are also being stored in shared_ptr, so that they are being double-deleted. I have not tracked down which object is being double deleted.

johnwason:
@Levi-Armstrong did you find the segfault source?

Levi-Armstrong:
I did not, I looked through but could not find anything that was doing anything questionable. It does move unique pointers into shared pointers but that shouldn't be an issue.

PR #334, opened 2023-06-17
Make abort configurable in ErrorTask

Levi-Armstrong:
@marip8 and @marrts  I am starting to think this should leverage a profile instead; what do you think? The use case is for when you want to spin up say a bunch of raster plans but you do not want to have a raster failure abort all of them because you will manage the failed one internal to the task, similar to the next best planner approach.

### DIFF
--- a/tesseract_task_composer/README.rst
+++ b/tesseract_task_composer/README.rst
@@ -454,6 +454,7 @@ The final task that is called in a task graph if error occurs
      class: ErrorTaskFactory
      config:
        conditional: false
+       trigger_abort: true
     
 Fix State Bounds Task
 ^^^^^^^^^^^^^^^^^^^^^
--- a/tesseract_task_composer/core/include/tesseract_task_composer/core/nodes/error_task.h
+++ b/tesseract_task_composer/core/include/tesseract_task_composer/core/nodes/error_task.h
@@ -57,6 +57,8 @@ protected:
   template <class Archive>
   void serialize(Archive& ar, const unsigned int version);  // NOLINT
 
+  bool trigger_abort_{ true };
+
   TaskComposerNodeInfo::UPtr runImpl(TaskComposerInput& input,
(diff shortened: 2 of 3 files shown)


marip8:
I'm not sure I would parameterize the error node like this. I think it might make more sense to modify the raster planner(s) to support best effort planning. It seems like you should be able to dynamically generate transition trajectories based on which rasters were successful

Levi-Armstrong:
I think the ability to configure the triggering an abort on failure will be needed for reusability of existing pipelines. Right now if I submit 10 raster if one fails the abort flag gets set and anyone that is not executing already will not execute. In this case, I want all to run to completion. It would be nice to see if we can come up with a way to connect graphs like conditional tasks.

marrts:
I feel like error and a widespread abort trigger should be two separate things. Which I think is kinda what you’re saying, but rather than making error configurable it might make more sense to add a new block that is a global abort. Then in something like the raster planner you could make it so that if a raster fails it could trigger a global abort outside of the actual raster plan. And that would make it easier to implement a best effort planner while maintaining readability.

Levi-Armstrong:
> I feel like error and a widespread abort trigger should be two separate things. Which I think is kinda what you’re saying, but rather than making error configurable it might make more sense to add a new block that is a global abort. Then in something like the raster planner you could make it so that if a raster fails it could trigger a global abort outside of the actual raster plan. And that would make it easier to implement a best effort planner while maintaining readability.

I agree, but for this to work we will need to figure out a good way to make TaskComposerGraph conditional to allow connecting. This was the same issue with using taskflow directly. The easies way is to make the TaskComposerGraph a task with internally spins up a dynamic task graph then responds to errors, but I think that would waist a thread. I will test with taskflow to see if that is the case.

Levi-Armstrong:
I will have a solution where a abort task is leverage when wanting to abort

PR #338, opened 2023-06-21
Added trajectory logger printout to trajectory checker

marrts:
This is essentially the same as https://github.com/tesseract-robotics/tesseract/pull/765, but for tesseract_planning


marrts:
@Levi-Armstrong, should these trajectory checking utils in tesseract_planning just make a `TrajArray` object and directly call the `checkTrajectory` method in tesseract_environment? Or should we still keep this separate? Right now it feels like duplicate code.

Levi-Armstrong:
Use composite version. I plan to move Tesseract Composite Instruction package to Tesseract repo and use it for joint trajectory and tool paths.

marrts:
> @marrts Is this ready to review?

Yes

Levi-Armstrong (approved):

PR #344, opened 2023-07-03
Fix TaskComposerServer segfault during tests

johnwason:
I found the cause of the segfault in `test_task_composer_core_unit.cpp`. When `TaskComposerServer` is destroyed, it destroys the `executors_`, `tasks_`, and `plugin_factory_` in some undefined order. The plugin factory must remain in scope until the other two or destroyed, or it will cause an error when the created objects are deleted. I added a destructor to make sure they are deleted in the correct order.  On Linux I guess it happened to be the right order by accident, but on Windows it was not.

This is potentially an issue in other parts of the codebase.


johnwason:
Please create another release if this passes.

Levi-Armstrong:
Looking at the standard the order at which member variables get destructed is reverse the order listed in the class. I created a PR which tests this which should avoid the need for a custom destructor.

Levi-Armstrong:
Replaced by #345

Levi-Armstrong:
@johnwason The fix as been applied and a new release has been created. Thanks for tracking down the issue.

johnwason:
@Levi-Armstrong  I am not sure I would trust the compiler to follow the standard... I wonder why it worked on Linux before? It should have done the same thing.

Levi-Armstrong:
I am not sure why it works on linux but might be more related to with how windows manages dynamically loaded libraries versus how Linux manages these.

PR #346, opened 2023-07-03
Add clone method to TaskComposerProblem

Levi-Armstrong:


Levi-Armstrong:
@johnwason It looks like windows builds are not using cache to allow faster building. Do you know if there is an issue in the CI file causing this?

johnwason:
I don't think I have caching enabled for windows. I have used `buildcache` on other projects for this purpose: https://github.com/robotraconteur/robotraconteur/blob/5660d41fe507aa29787fec3707b564975b935e52/.github/workflows/main.yml#L244

Levi-Armstrong:
I thought the previous versions of windows CI was caching vcpkgs so they did not have to be downloaded again?

johnwason:
Which run are you looking at? The most recent one I am seeing a cache hit.

Levi-Armstrong:
At one point the windows builds were as fast or faster than the ubuntu builds so not sure what has changed.

johnwason:
We may have disabled optimization on those builds to make it faster. I don't see anything out of the ordinary on the workflows. If there is a cache miss on vcpkg it will take a while. We can go back to using the ROS noetic based CI build if the vcpkg based one is too slow.

Levi-Armstrong:
I think it is fine, just wanted to check. It finishes around the same time as clang-tidy so not a huge issue.

PR #350, opened 2023-07-11
Reduce CI-built docker image size 

marip8:
Mirrors [`tesseract` PR 911](https://github.com/tesseract-robotics/tesseract/pull/911)


Levi-Armstrong:
So it looks like removing all -dev apt components breaks the build because dependencies required by tesseract are still required to build these packages. @marip8 Any ideas on how best to resolve this?

marip8:
> So it looks like removing all -dev apt components breaks the build because dependencies required by tesseract are still required to build these packages. @marip8 Any ideas on how best to resolve this

Addressed in [`tesseract` PR 913](https://github.com/tesseract-robotics/tesseract/pull/913)

marip8:
I'll wait ~30 minutes for the new `tesseract` docker image to be pushed before I re-run CI here. I'll remove the deletion of the `-dev` libraries from this CI config as well

marip8:
Looks like OMPL failed a unit test (seemingly randomly). @Levi-Armstrong any ideas why?

Levi-Armstrong (approved):

Issue #355, opened 2023-07-28
Is trajopt_ifopt support for CollisionEvaluatorType::CONTINUOUS and CollisionEvaluatorType::LVS_CONTINUOUS correctly implemented?


rjoomen:
I noticed a typo [here](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_motion_planners/trajopt_ifopt/src/trajopt_ifopt_utils.cpp#L305), where `LVSDiscreteCollision_` should be either `ContinuousCollision_` or `LVSContinuousCollision`.

But it's not clear whether this code even distinguishes between CONTINUOUS and LVS_CONTINUOUS. Is this a bug, or do I not understand how it works?

Levi-Armstrong:
It should be LVSContinuousCollision. We make a distinction between DISCRETE and LVS_DISCRETE because they involve different number of parameters. In the case of CONTINUOUS AND LVS_CONTINUOUS they both require two inputs, but it looks like we may not make a distinction between the two under the hood.

PR #357, opened 2023-08-01
Fix TrajOpt Ifopt collision cost and constraint naming

Levi-Armstrong:
Address issue #355


Levi-Armstrong:
The bug fix in trajopt repository is resulting in the glass up right example test failing. Will need to spend some time to figure out why it is failing.

Issue #358, opened 2023-08-11
RasterFtPipeline showing odd behavior 


peik-etzell:
I have been trying to do some planning with the RasterFtPipeline in tesseract_task_composer, and found some suspect behavior:

1. Transitions 1 and 2 are always completely broken, as if the _transition_ part of the trajectory was replaced by the $freespace_{start\rightarrow 1}$ / $freespace_{n \rightarrow end}$ motion
1. $Transition_{i \rightarrow i+1}$ seems to be queued for computation before $raster_i$ and $raster_{i+1}$ are finished computing, even though it is obviously dependent on the output of their computations (it seems the transition does wait for raster _i_, but I am not certain of anything here)
    - The freespace-task seems to plan to the output of the MinLengthTask as I understand it, which means there is a huge joint-interpolated jump in some cases
    - My testing has been purposefully _underconstrained_ (3-5 dof pose target, 9 dof robot), which makes this problem apparent, I suspect it might go unnoticed if the start- and end-states are fully constrained
1. _Sometimes_ the $freespace_{n \rightarrow end}$ motion is also bugged, jumping to one of the raster<>raster transitions

See the attached video, some notes:
- Transitions 1 and 2 broken
- Transition 3 plans without final target position -> joint-interpolated jump
- Transition 4 is perfect
- Transitions 5 and 6 do the same as T3, even though the jump is less noticable
- Transition 7 is again perfect

https://github.com/tesseract-robotics/tesseract_planning/assets/51529080/882dad5c-6211-4ac7-8d6f-e761beb375b0

I have made an as-minimal-as-possible [repo](https://github.com/peik-etzell/tesseract_raster_demo) that produces the above video for reproducibility.

Levi-Armstrong:
Than you for reporting and I will take look. Have you tried outputting the result dot graph?

Levi-Armstrong:
@marip8 and @marrts Have you been using the raster planners?

peik-etzell:
Ah yes I forgot to mention some observations from the dotgraph. I'll have to take a look at the dotgraph from this specifically later today, but testing our own robot I found that one transition subgraph was (always?) left white (uncomputed?), and this uncomputed one seemed to always be dependent on two [1] outputs from the rasters (dotted arrows), while all other transitions got one [0] and one [1], or two [0]. What I read in the source it is because the rasters are conditional, but this model seems incompatible with the 1 to many usage here?

Levi-Armstrong:
[...]

PR #360, opened 2023-08-14
Fix Raster and RasterOnly Tasks

Levi-Armstrong:
Address issue #358

### DIFF
--- a/tesseract_task_composer/core/include/tesseract_task_composer/core/nodes/sync_task.h
+++ b/tesseract_task_composer/core/include/tesseract_task_composer/core/nodes/sync_task.h
@@ -1,0 +1,78 @@
+/**
+ * @file sync_task.h
+ *
+ * @author Levi Armstrong
+ * @date August 13, 2023
+ * @version TODO
+ * @bug No known bugs
+ *
+ * @copyright Copyright (c) 2023, Plectix Robotics
+ *
+ * @par License
+ * Software License Agreement (Apache License)
+ * @par
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * @par
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef TESSERACT_TASK_COMPOSER_SYNC_TASK_H
+#define TESSERACT_TASK_COMPOSER_SYNC_TASK_H
+
+#include <tesseract_common/macros.h>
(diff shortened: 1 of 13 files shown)


tesseract_task_composer/core/include/tesseract_task_composer/core/nodes/sync_task.h
```diff
@@ -0,0 +1,73 @@
...
+
+namespace tesseract_planning
+{
+class TaskComposerPluginFactory;
+
+/**
+ * @brief This is used as a syncronization point in your task graph
+ * This task has no inputs or output and is not conditional
+ */
+class SyncTask : public TaskComposerTask
```
marip8, line 43: What is the point of this task if it doesn't actually do anything and doesn't take inputs, set outputs?

Levi-Armstrong: It is a synchronization point.

marrts:
You need to change [this](https://github.com/Levi-Armstrong/tesseract_planning/blob/be0de3166d954ed11e5f457e6671e518b359fde4/tesseract_task_composer/planning/src/nodes/raster_motion_task.cpp#L68) line because multiple types of tasks might have the same name. Potentially add an additional prefix for raster/transition/freespace as this function is called 3 separate times and the type of motion is known like [here](https://github.com/Levi-Armstrong/tesseract_planning/blob/be0de3166d954ed11e5f457e6671e518b359fde4/tesseract_task_composer/planning/src/nodes/raster_motion_task.cpp#L178).

tesseract_task_composer/core/include/tesseract_task_composer/core/nodes/sync_task.h
marrts: How exactly is it used? To make sure parallel tasks start at the same time or something? I don't fully understand why that is needed and I don't see it being used anywhere.

Levi-Armstrong: It would replace the StartTask as a general synchronization task.

marip8: How does it synchronize? I don't see anything in the class itself that makes sure all of its input tasks are complete before setting success. Is that something that taskflow is actually doing behind the scenes?

Levi-Armstrong: It is something that the executor will inherently do based on the graph structure but it needs a actual task so you can connect all parallel tasks to so when it gets to this task the executor will not proceed until all preceding tasks are finished.

marip8: Makes sense; we should add something like that comment to the doc string of the class. It wasn't totally clear to me why this should work. It's also a little unclear that not accepting input/output data does not mean it cannot accept input/output connections to other tasks

Levi-Armstrong: I updated the documentation of the class.

PR #362, opened 2023-08-17
Draft: Removed unused CMake option from task composer

marip8:


marip8:
After looking a little deeper, I see that this option is in fact used elsewhere in this package. It seems like a better approach would be to look for specific motion planner targets instead and make compile definitions based on which ones are available since they are all optionally compilable

Levi-Armstrong:
> After looking a little deeper, I see that this option is in fact used elsewhere in this package. It seems like a better approach would be to look for specific motion planner targets instead and make compile definitions based on which ones are available since they are all optionally compile-able.
> 
> We need to do this with the time parameterization algorithms as well since their compilation is also optional

I agree

marip8:
Closing; see #364 for a discussion on how to resolve this issue

PR #367, opened 2023-08-22
Draft: Profile Remapping

marip8:
Ports #255 to the master branch

## TODO
- [ ] Add commit for adding all default profiles to profile dictionary


marip8:
Instead of adding functions to add all the task composer and planner profiles, @Levi-Armstrong and I discussed giving each motion planner/task composer task a default profile member that could optionally be used if the requested profile does not exist in the profile dictionary

rjoomen:
Friendly ping.

rjoomen:
Can this be closed?

Levi-Armstrong:
@marip8 What are your thoughts?

marip8:
I think most of these proposed changes have already been addressed by the profile base class refactor, so we can close it. I captured comments about not loading default profiles in a new issue (#615 ) so we can address that separately in the near future

PR #370, opened 2023-08-24
Add new task composer plugin factory constructor

marip8:
This PR adds another constructor and `loadConfig` method to the task composer plugin factory to configure it with a `tesseract_common::TaskComposerPluginInfo` struct. This makes it easier to programmatically configure a plugin loader (e.g., from a GUI representation of a task composer graph/task)

### DIFF
--- a/.github/workflows/add_sources.sh
+++ b/.github/workflows/add_sources.sh
@@ -1,4 +1,5 @@
 #! /bin/bash
+apt update -qq
 apt install -y --no-install-recommends software-properties-common
 add-apt-repository -y ppa:ros-industrial/ppa
 apt update -qq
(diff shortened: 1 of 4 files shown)


marip8:
Not sure what happened to the taskflow library on Focal; the stable job fails because it can't install taskflow

Levi-Armstrong:
> Not sure what happened to the taskflow library on Focal; the stable job fails because it can't install taskflow

It should get it from here: https://launchpad.net/~ros-industrial/+archive/ubuntu/ppa/+packages

Levi-Armstrong:
Actually it can also get it from ros: https://repositories.ros.org/status_page/ros_noetic_default.html?q=taskflow

Levi-Armstrong:
I pushed a commit which I think fixed the CI build. The only thing left is unit tests before this can be merged.

Levi-Armstrong (approved):
Pending that the pipeline passes

PR #373, opened 2023-09-05
Update to leverage cmake components

Levi-Armstrong:


rjoomen:
Is the way to use these components already documented somewhere? I noticed for example [this comment](https://github.com/tesseract-robotics/tesseract_docs/issues/19#issuecomment-1706891512) where some useful ones are mentioned, but neither the readme, not tesseract_docs mentions them yet, it seems.

Levi-Armstrong:
I have not added to the docs yet, but need to figure out how best manage this going forward

PR #381, opened 2023-09-21
Added jammy build to CI

marip8:
Changes per commit message


Levi-Armstrong:
Should also add it to Unstable.

Levi-Armstrong (changes requested):
Should also add it to Unstable.

Levi-Armstrong (approved):

PR #383, opened 2023-09-25
Add user data support to CompositeInstruction

Levi-Armstrong:


marip8:
What is the intended purpose of user data in the composite instruction?

Levi-Armstrong:
> What is the intended purpose of user data in the composite instruction?

The composite instruction is not a poly type so this allow users to store additional data with the program if needed.

marip8:
Did you have any specific ideas of what types of data might be put here? I worry that just adding extra members to these classes to support any possible use-case will end up bloating them and making it hard to understand what purpose the members are supposed to be for.

Levi-Armstrong:
This is to prevent such a thing because the user data allows you to contain any additional data you may need to embed with the program without having to add anything additional to the composite instruction class. I don't understand why this would be hard for members to understand the purpose since the name is self explanatory (user data) and the use case is up to the user.

Levi-Armstrong:
Also this common practice, which can found in ogre, ogre2, gz-rendering, etc.

https://github.com/gazebosim/gz-rendering/blob/61bf635a5b6daf4918f37f837d0a5076adc5c879/include/gz/rendering/Node.hh#L37-L47

marip8:
That's fine; I just wanted to caution against adding arbitrary-looking fields to classes to accommodate theoretical use-cases. It's easy to just keep adding members to classes to support new use-cases rather than thinking harder about a potentially better solution that uses the existing framework

PR #394, opened 2023-10-11
Unused includes cleanup

rjoomen:
Used clangd to check all files for unused includes

### DIFF
--- a/tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
+++ b/tesseract_command_language/include/tesseract_command_language/joint_waypoint.h
@@ -29,7 +29,6 @@
 #include <tesseract_common/macros.h>
 TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <Eigen/Core>
-#include <memory>
 #include <vector>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
--- a/tesseract_command_language/include/tesseract_command_language/move_instruction.h
+++ b/tesseract_command_language/include/tesseract_command_language/move_instruction.h
@@ -28,7 +28,6 @@
 
 #include <tesseract_common/macros.h>
 TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
-#include <vector>
 #include <Eigen/Geometry>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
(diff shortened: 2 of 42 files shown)


rjoomen:
Build failures for stable are because of the latest trajopt_ifopt collision refactor

Levi-Armstrong:
Restarted CodeCov CI because it failed for a weird reason.

Levi-Armstrong (approved):
LGTM, good to merge when CI passes

marip8:
> Build failures for stable are because of the latest trajopt_ifopt collision refactor

We should update the `dependencies.repos` file to point to the correct version of `trajopt` to fix the stable builds before merging

rjoomen:
> > Build failures for stable are because of the latest trajopt_ifopt collision refactor
> 
> We should update the `dependencies.repos` file to point to the correct version of `trajopt` to fix the stable builds before merging. We can either point to commit `2d965b9` or create a new tagged version (`0.7.0` I think)

But this is unstable<->unstable, no need to fix the references of stable before a release is made, right?

marip8:
I think see what you're saying, but I think my definitions of stable and unstable may be slightly different. Here are my expectations for PRs:

- The current state of the code (i.e., a PR) should *always pass a set of CI tests for a strictly defined set of dependencies (i.e., tagged versions (not branch names) of dependency repositories that are guaranteed to result in a successful build of this repo: `dependencies.repos`)
   - This is what I think of as a "stable" build
   - *always, of course there are justifiable exceptions to this rule
- The current state of the code (i.e., a PR) doesn't necessarily have to pass a set of CI tests for a more loosely defined set of dependencies (i.e., branch names of dependency repositories: `dependencies_unstable.repos`)
  - This is what I think of as an "unstable" build and are denoted in this repo's CI as "Unstable"
  - The only value I see in these tests is that it tells whether this repo must also change at some point in the future when we decide to officially support the new version of the dependency.

For this PR, we are choosing to update this repository such that it supports a new version of `trajopt`. Therefore, I think we ought to change the "stable" dependencies to reflect that. As a user, I don't want to pull this repo and the dependencies in `dependencies.repos` only to figure out that the build fails because the dependencies aren't specified correctly.

I'm not sure that I see this repository itself as being stable or unstable. To me stability is about whether or not a particular version of this repository builds successfully with what that version calls its "stable" dependencies (i.e. `dependencies.repos`)

This is probably not the right place to go into this level of discussion about CI practices, but we should document somewhere what are our expectations about PRs in terms of CI and what are our definitions of stable and unstable

Levi-Armstrong:
The current definitions used within Tesseract repositories
- Stable CI is ran using the latest release dependencies versions to detect breaking changes made in this repository or dependency repository.
- Unstable CI is ran using the latest for identified dependencies repositories which is were new development happens until the next release.

Issue #399, opened 2023-10-23
Remove planning task composer `input_instruction`


marip8:
The `PlanningTaskComposerProblem` class has a member called `input_instruction` that presumably holds the instruction that the first task is supposed to use as input.

https://github.com/tesseract-robotics/tesseract_planning/blob/d2b11debc8b6ab414c078333430db4004771445b/tesseract_task_composer/planning/include/tesseract_task_composer/planning/planning_task_composer_problem.h#L90

This seems unecessarily redundant because each task, including the first task, can specify an input key in the `DataStorage` object from which to extract its relevant input. We should consider removing the `input_instruction` field in favor of using the `DataStorage` object. I have tested this in the [SNP repo here](https://github.com/marip8/scan_n_plan_workshop/blob/de58d8df7969fd2dfdbeaf0958134f8a9824120a/snp_motion_planning/src/planning_server.cpp#L506-L509) and it works as expected.

A quick code search shows that only the [process planning input task](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_task_composer/planning/src/nodes/process_planning_input_task.cpp#L69) and a unit test seem to use this member anyway.

Levi-Armstrong:
This decision was to have a clear distinction between the input to the planning pipeline and the data generated as part of the pipeline. I want the problem to hold all required data to initiated a pipeline required for logging/serialization. This way you only have to serialize and log the problem to allow recreating failed or undesirable plans.

Levi-Armstrong:
This also has other nice properties from the Tesseract Studio perspective for creating universal debug and planning tooling.

marip8:
> I want the problem to hold all required data to initiated a pipeline required for logging/serialization

I don't think this is realistically achievable with only the `input_instruction` field. There could be tasks in any part of the pipeline that want to look up pre-populated data in the `DataStorage` class. Also the first task may want some type of data that is not an `Instruction` type.

Levi-Armstrong:
[...]

PR #407, opened 2023-11-06
Move TaskComposerProblem input to base class and change type to tesseract_common::AnyPoly

Levi-Armstrong:
Addresses #399


tesseract_task_composer/planning/src/nodes/process_planning_input_task.cpp
marip8, line 1: Since all tasks will now be required to put their input data in the problem directly, would it make more since for the executor to put the input data in the correct place in the internal data storage object, rather than having this separate task do it? Otherwise, it seems like all pipelines from here on out will need to have this as the first task in the pipeline

Levi-Armstrong: Lets add it as an issue to investigate if this is the right thing to do.

PR #411, opened 2023-11-08
Update dependencies.repos

Levi-Armstrong:


Levi-Armstrong:
@marip8 I think your changes to CI has broken the repos CI builds. I am trying to get a release out today do you know how to fix this?

``` bash
'init' returned with code '0' after 0 min 7 sec
UNDERLAY '/root/tesseract-robotics/tesseract_target_ws/install' does not contain a setup.bash
```

marip8:
The `/opt` directory in the built docker image has been cleared, so none of the build artifacts exist in it. I think it may be something with the `AFTER_SCRIPT` step that tries to remove the build directory. I'm going to run locally now to figure this out

Levi-Armstrong:
I reran it and it now shows this error message. 
```
'init' returned with code '0' after 0 min 6 sec
UNDERLAY '${BASEDIR}/tesseract-robotics/tesseract_target_ws/install' does not contain a setup.bash
```

marip8:
Running it locally (for `tesseract`, not `tesseract_planning`) shows that things build in the `/opt` directory correctly and the unit tests run, but for some reason none of that gets saved into the docker image when it is committed

marip8:
Per ICI:

```
BASEDIR (default: $HOME): Base directory in which the upstream, target, and downstream workspaces will be built. Note: this directory is bind-mounted, so it can be read by the CI service, but its contents will not persist in the image configured by DOCKER_COMMIT
```

For some reason this is not true when `BASEDIR=/root` because the build artifacts remain in `/root`, but this does seem to be true with `/opt`. I'm running a test now with `BASEDIR=/home` to see if the same thing happens. If it does, I think the solution would be to set `BASEDIR=/home` and use `AFTER_SCRIPT` to copy the build artifacts in the install directories to `/opt`

marip8:
Re-running CI now that the `tesseract` docker images have been regenerated

marip8:
I just pushed a `0.21.1` release to `tesseract` to regenerate the tagged docker images for the stable builds. We can re-run those failed jobs after that finishes

PR #403, opened 2023-11-03
Add toleranced waypoints to TrajOpt Solver

marrts:
In conjunction with [this PR](https://github.com/tesseract-robotics/trajopt/pull/354) in TrajOpt. This allows users to set tolerances on Cartesian and joint waypoints. Additionally you can now set both costs and constraints for waypoints so you can drive the solver to match the waypoint exactly while enforcing strict limits that are nonzero on where the waypoint can actually move.

This is a breaking change because it changes the interface for the TrajOpt default plan profile.

I think this new interface would also be useful for the TrajOpt IFOPT work that is going on.

PR #427, opened 2023-12-14
Add cartesian waypoints to fixed indices

rjoomen:
I'm not entirely sure if this is correct, in my testing it reduced planning success, but improved planning speed.


Levi-Armstrong:
I would recommend see if we can leverage the trajopt profile to indicate of a waypoint should be treated as fixed. In case when you have extra degrees of freedom these changes will not be ideal.

rjoomen:
Sure, I was just trying to fix the todo. How exactly do you suggest I do that? And does that only apply to the Cartesian waypoints or also to the other types?

Also, the joint and state waypoints still have todos saying they should not be fixed if they are of term type cost, but I wasn't sure how to obtain the correct profile information.

Levi-Armstrong:
I would probably update the TrajOpt Plan Profile to return a bool to indicate if it should be treated as fixed. If we think there may be other things that we want to return we could create a struct which gets return allowing us to expand it in the future if needed.

rjoomen:
How about this?

(I am unsure about how to handle the [state waypoint](https://github.com/rjoomen/tesseract_planning/blob/cffb81493e545fac24967b7d2ce6df469d433219/tesseract_motion_planners/trajopt/src/trajopt_motion_planner.cpp#L310C23-L310C23) type, btw)

rjoomen:
> Does this change mean that if we have a fully constrained Cartesian waypoint and a seed joint state for that waypoint the output will always be equivalent to the seed? It looks like trajopt just skips the fixed steps and doesn't add them to the problem at all. This would be problematic for applications where you have redundant axes that you want TrajOpt to smooth across waypoints, but you still have strict Cartesian requirements

That is a good point. @Levi-Armstrong should cartesian waypoints maybe never be fixed (which effectively is the current situation)? Should I remove that part and only leave the logic pertaining the joint waypoints?

rjoomen:
I went ahead and removed the fixed cartesian waypoint logic. Now this PR only fixes [this](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_motion_planners/trajopt/src/trajopt_motion_planner.cpp#L288C32-L288C32) and [this](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_motion_planners/trajopt_ifopt/src/trajopt_ifopt_motion_planner.cpp#L313) todo.

Now how should [state waypoints](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_motion_planners/trajopt/src/trajopt_motion_planner.cpp#L308) be handled?

Levi-Armstrong:
> I went ahead and removed the fixed cartesian waypoint logic. Now this PR only fixes [this](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_motion_planners/trajopt/src/trajopt_motion_planner.cpp#L288C32-L288C32) and [this](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_motion_planners/trajopt_ifopt/src/trajopt_ifopt_motion_planner.cpp#L313) todo.
> 
> Now how should [state waypoints](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_motion_planners/trajopt/src/trajopt_motion_planner.cpp#L308) be handled?

State waypoints should be treated as fixed joint state.

rjoomen:
> State waypoints should be treated as fixed joint state.

But the todo suggests to not make them fixed if they are cost instead of constraint.

marrts:
> State waypoints should be treated as fixed joint state.

So are state waypoints assumed to be always fixed in regards to their position? They aren't fixed in velocity, acceleration, or time stamp because time parameterization tasks expect state waypoints as inputs. So to me it seems arbitrary that you'd expect TrajOpt to treat state waypoints position values as fixed. I know we also have joint waypoints that could or couldn't be fixed, but I feel like there's probably a scenario that a previous planner produces a state waypoint that you want TrajOpt to treat as only a cost.

I also think the fact that for a given profile we have to set all types of waypoints to either a cost or constraint makes this more complicated. That's part of the motivation behind my other PR, #403, where I changed the ` TrajOptDefaultPlanProfile` to have different configs for each waypoint type. I could add another one for state waypoints too, giving control over setting it to a cost (not fixed) vs constraint (fixed).
https://github.com/tesseract-robotics/tesseract_planning/blob/d2fa313f47f05aa909e639cac12d1fe62c0586ce/tesseract_motion_planners/trajopt/include/tesseract_motion_planners/trajopt/trajopt_waypoint_config.h#L43 
https://github.com/tesseract-robotics/tesseract_planning/blob/d2fa313f47f05aa909e639cac12d1fe62c0586ce/tesseract_motion_planners/trajopt/include/tesseract_motion_planners/trajopt/trajopt_waypoint_config.h#L71
https://github.com/tesseract-robotics/tesseract_planning/blob/d2fa313f47f05aa909e639cac12d1fe62c0586ce/tesseract_motion_planners/trajopt/include/tesseract_motion_planners/trajopt/profile/trajopt_default_plan_profile.h#L55-L58

tesseract_motion_planners/trajopt_ifopt/src/profile/trajopt_ifopt_default_plan_profile.cpp
```diff
@@ -127,6 +127,13 @@ void TrajOptIfoptDefaultPlanProfile::apply(TrajOptIfoptProblem& problem,
   }
 }
 
+bool TrajOptIfoptDefaultPlanProfile::isFixedJoint() const
+{
+  // If the term type is constraint and all coefficients are non-zero
+  return (term_type == TrajOptIfoptTermType::CONSTRAINT) &&
+         (abs(joint_coeff.array()) >= std::numeric_limits<double>::epsilon()).all();
```
Levi-Armstrong, line 134: @rjoomen Why do all coeff have to be non zero?

Levi-Armstrong:
I believe this has been addressed by #718

PR #433, opened 2024-01-08
Add optional namespace field to task nodes

marrts:
Currently there is a  property in `tesseract_task_composer` that makes it difficult to add multiples of the same task in a pipeline. This PR addresses that issue by adding an optional `"namespace"` (we can maybe pick a better name, but this is what the `addProfile` method calls it) field to the task yaml object.

Going from:
```yaml
Task:
  class: TaskFactory
  config:
    ...
```
To:
```yaml
Task:
  class: TaskFactory
  namespace: TaskNS #(optional)
  config:
    ...
```

When adding a profile of a planner you need to specify a namespace, profile_name, and a shared_ptr to the profile definition. (see: [here](https://github.com/tesseract-robotics/tesseract_planning/blob/2e1e31c3899469280156f9630cae544e54fed6ea/tesseract_command_language/include/tesseract_command_language/profile_dictionary.h#L122))
```C++
void addProfile(const std::string& ns, const std::string& profile_name, std::shared_ptr<const ProfileType> profile)
```

## Why is this necessary?
Imagine a scenario that you want to do a freespace planner that tries to do the lowest effort planner first and then progressively goes to more robust planners as things fail. In this graph you might have multiple TrajOpt and/or Contact Check steps that you want to all be configured in the same way. When you get to a given task for a given waypoint a [2D lookup is performed](https://github.com/tesseract-robotics/tesseract_planning/blob/2e1e31c3899469280156f9630cae544e54fed6ea/tesseract_command_language/include/tesseract_command_language/profile_dictionary.h#L190) to find the correct profile to be used. The task namespace and the name of the profile at the waypoint, then the profile object is pulled and applied to set up the actual planner used. Currently when creating a yaml file to set up a taskflow the namespace field is automatically generated from the task name as determined by the key in the yaml file. For example, suppose you define a task like this:
``` yaml
OMPLMotionPlannerTask:
  class: OMPLMotionPlannerTaskFactory
  config:
    conditional: true
    inputs: [output_data]
    outputs: [output_data]
    format_result_as_input: true
```
Here the `namespace` of the task will be `OMPLMotionPlannerTask`, but you are limited to only one task with this namespace. In my C++ code I'll make a definition of this profile so tesseract knows what to do whenever it comes across an OMPLMotionPlannerTask in a taskflow process. That will be defined by something like:
```C++
profile_dict->addProfile<tesseract_planning::OMPLPlanProfile>("OMPLMotionPlannerTask", PROFILE, profile);
```
In my above described scenario I need to do something that has multiple TrajOpt and multiple Contact Check tasks, so I need to set up my yaml file with something like:
``` yaml
TrajOptMotionPlannerTask1:
  class: TrajOptMotionPlannerTaskFactory
  config:
    conditional: true
    inputs: [output_data]
    outputs: [output_data]
    format_result_as_input: false
DiscreteContactCheckTask1:
  class: DiscreteContactCheckTaskFactory
  config:
    conditional: true
    inputs: [output_data]
TrajOptMotionPlannerTask2:
  class: TrajOptMotionPlannerTaskFactory
  config:
    conditional: true
    inputs: [output_data]
    outputs: [output_data]
    format_result_as_input: false
DiscreteContactCheckTask2:
  class: DiscreteContactCheckTaskFactory
  config:
    conditional: true
    inputs: [output_data]
```
This means in my C++ code I'll need to set up profiles for each of these:
```C++
// Add for first instance
profile_dict->addProfile<tesseract_planning::ContactCheckProfile>(
            "DiscreteContactCheckTask1", PROFILE,
            std::make_shared<tesseract_planning::ContactCheckProfile>(contact_check_lvs, contact_check_dist));
profile_dict->addProfile<tesseract_planning::TrajOptPlanProfile>("TrajOptMotionPlannerTask1", PROFILE,
                                                                         createTrajOptToolZFreePlanProfile());
profile_dict->addProfile<tesseract_planning::TrajOptCompositeProfile>("TrajOptMotionPlannerTask1", PROFILE,
                                                                      createTrajOptProfile());
                                                                      
// Add for first instance of task                                                                      
profile_dict->addProfile<tesseract_planning::ContactCheckProfile>(
            "DiscreteContactCheckTask2", PROFILE,
            std::make_shared<tesseract_planning::ContactCheckProfile>(contact_check_lvs, contact_check_dist));
profile_dict->addProfile<tesseract_planning::TrajOptPlanProfile>("TrajOptMotionPlannerTask2", PROFILE,
                                                                         createTrajOptToolZFreePlanProfile());
profile_dict->addProfile<tesseract_planning::TrajOptCompositeProfile>("TrajOptMotionPlannerTask2", PROFILE,
                                                                      createTrajOptProfile());
                                                                      
```
This can get really tedious and cumbersome if you want to be doing a lot of modifications to your pipeline, especially as you get to increasingly complex pipelines, forcing you to recompile code.

This proposed change simplifies this process on the C++ side by adding another, optional, field to the yaml constructor that specifies the namespace, turning my scenario into this (note there is no effect on how edges are made:
```yaml
TrajOptMotionPlannerTask1:
  class: TrajOptMotionPlannerTaskFactory
  namespace: TrajOptMotionPlannerTask
  config:
    conditional: true
    inputs: [output_data]
    outputs: [output_data]
    format_result_as_input: false
DiscreteContactCheckTask1:
  class: DiscreteContactCheckTaskFactory
  namespace: DiscreteContactCheckTask
  config:
    conditional: true
    inputs: [output_data]
TrajOptMotionPlannerTask2:
  class: TrajOptMotionPlannerTaskFactory
  namespace: TrajOptMotionPlannerTask
  config:
    conditional: true
    inputs: [output_data]
    outputs: [output_data]
    format_result_as_input: false
DiscreteContactCheckTask2:
  class: DiscreteContactCheckTaskFactory
  namespace: DiscreteContactCheckTask
  config:
    conditional: true
    inputs: [output_data]
```
If a `namespace` field is detected, then the task namespace points to this string rather than the yaml key for the task. The following C++ code would then be sufficient:
```C++
// Add only once
profile_dict->addProfile<tesseract_planning::ContactCheckProfile>(
            "DiscreteContactCheckTask", PROFILE,
            std::make_shared<tesseract_planning::ContactCheckProfile>(contact_check_lvs, contact_check_dist));
profile_dict->addProfile<tesseract_planning::TrajOptPlanProfile>("TrajOptMotionPlannerTask", PROFILE,
                                                                         createTrajOptToolZFreePlanProfile());
profile_dict->addProfile<tesseract_planning::TrajOptCompositeProfile>("TrajOptMotionPlannerTask", PROFILE,
                                                                      createTrajOptProfile());
```
I can then add as many of these tasks to my pipeline without the need to recompile.

## Why can't we just assign all tasks in a graph the same namespace?
It might seems like this unnecessarily over-complicates things. If we want to re-use a task of the same type then maybe we should assume all tasks of the same type in the same graph have the same namepsace, but this is not always true. You could have a scenario where you have multiple tasks of the same planner type that you want to try different parameters on. A common one I use that fits this description is water-falling Descartes from sparse sampling to dense sampling if I know I need to do a lot of motion plans where _most_ will work with sparse sampling, but some require a denser sampling. Your resulting graph might look something like this:
image
In this scenario you want each Descartes solver to use a different profile.

## Where's the catch?
There is one complication with my changes that I'm not a huge fan of. Here is an example of a graph with this change where it tries just TrajOpt first and then falls back to OMPL if that fails (Note I think there are potentially some issues with this pipeline, but it's just for demonstration sake)
image
The downside here is that you can see we have multiple diamond blocks with the same name, making debugging your graph potentially more difficult if you have a poorly formed YAML file. In the following image I've made an intentional typo where I pointed the second TrajOpt step to the first Contact Check instead of the second one, making the graph very confusing and hard to debug:
image
Here is that same graph without the namespace change (it seems marginally easier to debug):
image

I would like other people to weigh in on what they feel is the best approach.

### DIFF
--- a/tesseract_motion_planners/core/include/tesseract_motion_planners/core/planner.h
+++ b/tesseract_motion_planners/core/include/tesseract_motion_planners/core/planner.h
@@ -37,7 +37,7 @@ public:
   using ConstPtr = std::shared_ptr<const MotionPlanner>;
   /** @brief Construct a basic planner */
   MotionPlanner() = default;
-  MotionPlanner(std::string ns);
+  MotionPlanner(std::string name);
   virtual ~MotionPlanner() = default;
   MotionPlanner(const MotionPlanner&) = delete;
   MotionPlanner& operator=(const MotionPlanner&) = delete;
@@ -45,10 +45,10 @@ public:
   MotionPlanner& operator=(MotionPlanner&&) = delete;
 
   /**
-   *  @brief Get the namespace of this planner
+   *  @brief Get the name of this planner
    *  @details This is also used as the namespace for the profiles in the profile dictionary
    */
-  const std::string& getNamespace() const;
+  const std::string& getName() const;
 
   /**
    * @brief Solve the planner request problem
@@ -79,7 +79,7 @@ public:
                              bool format_result_as_input);
 
 protected:
-  std::string ns_;
+  std::string name_;
 };
 }  // namespace tesseract_planning
 #endif  // TESSERACT_PLANNING_PLANNER_H
--- a/tesseract_motion_planners/core/include/tesseract_motion_planners/core/planner.h
+++ b/tesseract_motion_planners/core/include/tesseract_motion_planners/core/planner.h
@@ -37,7 +37,7 @@ public:
   using ConstPtr = std::shared_ptr<const MotionPlanner>;
   /** @brief Construct a basic planner */
   MotionPlanner() = default;
-  MotionPlanner(std::string name);
+  MotionPlanner(std::string ns);
   virtual ~MotionPlanner() = default;
   MotionPlanner(const MotionPlanner&) = delete;
   MotionPlanner& operator=(const MotionPlanner&) = delete;
@@ -45,10 +45,10 @@ public:
   MotionPlanner& operator=(MotionPlanner&&) = delete;
 
   /**
-   *  @brief Get the name of this planner
+   *  @brief Get the namespace of this planner
    *  @details This is also used as the namespace for the profiles in the profile dictionary
    */
-  const std::string& getName() const;
+  const std::string& getNamespace() const;
 
   /**
    * @brief Solve the planner request problem
@@ -79,7 +79,7 @@ public:
                              bool format_result_as_input);
 
 protected:
-  std::string name_;
+  std::string ns_;
 };
 }  // namespace tesseract_planning
 #endif  // TESSERACT_PLANNING_PLANNER_H
--- a/tesseract_task_composer/core/src/task_composer_node.cpp
+++ b/tesseract_task_composer/core/src/task_composer_node.cpp
@@ -52,7 +52,7 @@ TaskComposerNode::TaskComposerNode(std::string name, TaskComposerNodeType type, 
 {
   try
   {
-    ns_ = config["namespace"] ? config["namespace"].as<std::string>() : name_;
+    ns_ = config["namespace"].IsDefined() ? config["namespace"].as<std::string>() : name_;
 
     if (YAML::Node n = config["conditional"])
       conditional_ = n.as<bool>();
--- a/tesseract_task_composer/core/src/task_composer_node.cpp
+++ b/tesseract_task_composer/core/src/task_composer_node.cpp
@@ -39,6 +39,7 @@ namespace tesseract_planning
 {
 TaskComposerNode::TaskComposerNode(std::string name, TaskComposerNodeType type, bool conditional)
   : name_(std::move(name))
+  , ns_(name_)
   , type_(type)
   , uuid_(boost::uuids::random_generator()())
   , uuid_str_(boost::uuids::to_string(uuid_))
--- a/tesseract_task_composer/core/src/task_composer_node.cpp
+++ b/tesseract_task_composer/core/src/task_composer_node.cpp
@@ -52,10 +52,7 @@ TaskComposerNode::TaskComposerNode(std::string name, TaskComposerNodeType type, 
 {
   try
   {
-    if (YAML::Node n = config["namespace"])
-      ns_ = n.as<std::string>();
-    else
-      ns_ = name_;
+    ns_ = config["namespace"] ? config["namespace"].as<std::string>() : name_;
 
     if (YAML::Node n = config["conditional"])
       conditional_ = n.as<bool>();
--- a/tesseract_task_composer/core/src/task_composer_node.cpp
+++ b/tesseract_task_composer/core/src/task_composer_node.cpp
@@ -51,6 +51,11 @@ TaskComposerNode::TaskComposerNode(std::string name, TaskComposerNodeType type, 
 {
   try
   {
+    if (YAML::Node n = config["namespace"])
+      ns_ = n.as<std::string>();
+    else
+      ns_ = name_;
+
     if (YAML::Node n = config["conditional"])
       conditional_ = n.as<bool>();
 
@@ -78,6 +83,9 @@ TaskComposerNode::TaskComposerNode(std::string name, TaskComposerNodeType type, 
 
 void TaskComposerNode::setName(const std::string& name) { name_ = name; }
 const std::string& TaskComposerNode::getName() const { return name_; }
+
+void TaskComposerNode::setNamespace(const std::string& ns) { ns_ = ns; }
+const std::string& TaskComposerNode::getNamespace() const { return ns_; }
 
 TaskComposerNodeType TaskComposerNode::getType() const { return type_; }
 
@@ -134,7 +142,8 @@ std::string TaskComposerNode::dump(std::ostream& os,
 
   if (conditional_)
   {
-    os << std::endl << tmp << " [shape=diamond, label=\"" << name_ << "\\n(" << uuid_str_ << ")";
+    os << std::endl << tmp << " [shape=diamond, label=\"" << name_ << "\\n";
+    os << "Namespace: " << ns_ << "\\n(" << uuid_str_ << ")";
 
     os << "\\n Inputs: [";
     for (std::size_t i = 0; i < input_keys_.size(); ++i)
--- a/tesseract_task_composer/taskflow/src/taskflow_task_composer_executor.cpp
+++ b/tesseract_task_composer/taskflow/src/taskflow_task_composer_executor.cpp
@@ -82,7 +82,7 @@ void TaskflowTaskComposerExecutor::removeFuture(const boost::uuids::uuid& uuid)
 TaskComposerFuture::UPtr TaskflowTaskComposerExecutor::run(const TaskComposerNode& node,
                                                            TaskComposerContext::Ptr context)
 {
-  auto taskflow = std::make_unique<tf::Taskflow>(node.getNamespace());
+  auto taskflow = std::make_unique<tf::Taskflow>(node.getName());
   if (node.getType() == TaskComposerNodeType::TASK)
     convertToTaskflow(static_cast<const TaskComposerTask&>(node), *context, *this, taskflow.get());
   else if (node.getType() == TaskComposerNodeType::PIPELINE)
--- a/tesseract_task_composer/taskflow/src/taskflow_task_composer_executor.cpp
+++ b/tesseract_task_composer/taskflow/src/taskflow_task_composer_executor.cpp
@@ -82,7 +82,7 @@ void TaskflowTaskComposerExecutor::removeFuture(const boost::uuids::uuid& uuid)
 TaskComposerFuture::UPtr TaskflowTaskComposerExecutor::run(const TaskComposerNode& node,
                                                            TaskComposerContext::Ptr context)
 {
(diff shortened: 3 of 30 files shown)


rjoomen:
This seems like a useful change. Decoupling node names from profile namespaces also allows for more descriptive naming.

The catch with the duplicate names in the dot graph could fixed by showing the node name instead of the namespace, or maybe better show both (although that might crowd the figures). So then the graph nodes would show:

```
TrajOptMotionPlannerTask1
ns: TrajoptMotionPlannerTask
(uuid)
Inputs: [output_data]
Outputs: [output_data]
```

Levi-Armstrong:
I think it is imported that the names be unique, but to accomplish the same thing you could update TaskComposerNode class to store the namespace and provide setters and getters. Then update the yaml parsing for the TaskComposerNode to extract the namespace if it exists otherwise use the name of the task. Update the dot to show the namespace if different. Then update the task to use the getter for the namespace instead of the name. @marrts What do you think?

marrts:
> I think it is imported that the names be unique, but to accomplish the same thing you could update TaskComposerNode class to store the namespace and provide setters and getters. Then update the yaml parsing for the TaskComposerNode to extract the namespace if it exists otherwise use the name of the task. Update the dot to show the namespace if different. Then update the task to use the getter for the namespace instead of the name when retrieving profiles. @marrts What do you think?

Okay, I'll work on this today. I agree that makes the most sense.

marrts:
In starting to do this I am realizing this goes from a fully-backwards compatible change to a breaking change since I have to change the constructors of all the tasks to have a namespace field. @Levi-Armstrong, is this still what you want me to do?

Levi-Armstrong:
> In starting to do this I am realizing this goes from a fully-backwards compatible change to a breaking change since I have to change the constructors of all the tasks to have a namespace field. @Levi-Armstrong, is this still what you want me to do?

That should not be necessary if you make the namespace part of the config.

``` yaml
TrajOptMotionPlannerTask2:
  class: TrajOptMotionPlannerTaskFactory
  config:
    namespace: TrajOptMotionPlannerTask
    conditional: true
    inputs: [output_data]
    outputs: [output_data]
    format_result_as_input: false
```

marrts:
Changed to work with namespace in the config. I'm not sure how we want to handle objects that inherit from `tesseract_planning::MotionPlanner`. I changed them all to use `ns_` instead of `name_` to match with the convention for getting profiles now used by all objects that inherit from `TaskComposerTask`, but anyone that has their own custom tasks will have to change to match this convention (although that is true for tasks that inherit from either `MotionPlanner` or `TaskComposerTask`)

Here is what it looks like in a dot graph now:
image

tesseract_motion_planners/core/include/tesseract_motion_planners/core/planner.h
```diff
@@ -79,7 +79,7 @@ class MotionPlanner
                              bool format_result_as_input);
 
 protected:
-  std::string name_;
+  std::string ns_;
 };
 }  // namespace tesseract_planning
```
Levi-Armstrong, line 84: Although at the task composer level we will be using the namespace as the name, I still think this should remain the name.

Levi-Armstrong (changes requested):

tesseract_motion_planners/core/include/tesseract_motion_planners/core/planner.h
marrts: I agree, changing `name_` to `ns_` feels weird. The reason I changed this is because it makes getting profiles feel inconsistent otherwise. 

By keeping `name_` In the `MotionPlanner` base class object, in all inherited Tasks, like Descartes or OMPL or TrajOpt, you'd have a line like this:
```C++
auto cur_plan_profile = getProfile<DescartesPlanProfile<FloatType>>(
  name_, profile, *request.profiles, std::make_shared<DescartesDefaultPlanProfile<FloatType>>());
```
But in something like `MinLengthTask` that inherits from `TaskComposerTask` you have:
```C++
auto cur_composite_profile =
      getProfile<MinLengthProfile>(ns_, profile, *problem.profiles, std::make_shared<MinLengthProfile>());
```
In one example you reference the `name_` and in the other you reference `ns_`, and unless you know the history of all of this that might be really confusing.


The other idea I had was to just make a method in the `MotionPlanner` base class and the `TaskComposerTask` base class that would be something like
```C++
template <typename ProfileType>
td::shared_ptr<const ProfileType> getProfile(const std::string& profile, const ProfileDictionary& profile_dictionary)
{
  return getProfile<ProfileType>(<ns_ or name_>, profile, profile_dictionary, std::make_shared<ProfileType>());
}
```
Then in both `MotionPlanner` and `TaskComposerTask` inherited classes you can just call
```C++
getProfile<DescartesPlanProfile<FloatType>>(profile, *request.profiles);
```
or
```C++
getProfile<MinLengthProfile>(profile, *problem.profiles);
```
And then we can replace all instances of getting the profile with a less confusing interface, and I can revert `ns_` back to `name_`

tesseract_task_composer/taskflow/src/taskflow_task_composer_executor.cpp
```diff
@@ -82,7 +82,7 @@ void TaskflowTaskComposerExecutor::removeFuture(const boost::uuids::uuid& uuid)
 TaskComposerFuture::UPtr TaskflowTaskComposerExecutor::run(const TaskComposerNode& node,
                                                            TaskComposerContext::Ptr context)
 {
-  auto taskflow = std::make_unique<tf::Taskflow>(node.getName());
+  auto taskflow = std::make_unique<tf::Taskflow>(node.getNamespace());
```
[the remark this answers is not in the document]
marrts: Yes, you are right

tesseract_motion_planners/core/include/tesseract_motion_planners/core/planner.h
Levi-Armstrong: I don't think all of that is necessary, because they two different packages that use profiles. One way could be to just document why motion planners are using its name. Another option is to update PlannerRequest to include the profile namespace parameter. Then in the arm planner have something like.

```
const std::string profile_ns = (request.profile_ns.empty() ? name_ : request.profile_ns;
```

marrts: Okay, I reverted the change of `name_` to `ns_`.

> `const std::string profile_ns = (request.profile_ns.empty() ? name_ : request.profile_ns;`

I think that might make it more confusing.

Levi-Armstrong:
@marrts can you also updated the readme to document the namespace where it seems appropriate?

marrts:
> @marrts can you also updated the readme to document the namespace where it seems appropriate?

I added the namespace field everywhere it fit and marked it as optional. I also added an example of leveraging it to reuse a task configuration.

tesseract_task_composer/core/src/task_composer_node.cpp
```diff
@@ -52,7 +52,7 @@ TaskComposerNode::TaskComposerNode(std::string name, TaskComposerNodeType type,
 {
   try
   {
-    ns_ = config["namespace"] ? config["namespace"].as<std::string>() : name_;
+    ns_ = config["namespace"].IsDefined() ? config["namespace"].as<std::string>() : name_;
```
Levi-Armstrong, line 55: Does this not work?
```
ns_ = (YAML::Node n = config["namespace"]) ? n.as<std::string>() : name_;
```

marrts: It gave me a clang_tidy warning for that:
```bash
 /__w/tesseract_planning/tesseract_planning/target_ws/src/tesseract_task_composer/core/src/task_composer_node.cpp:55:11: error: implicit conversion 'YAML::detail::unspecified_bool_type' (aka 'void (*)(unspecified_bool::NOT_ALLOWED *)') -> bool [readability-implicit-bool-conversion,-warnings-as-errors]
    ns_ = config["namespace"] ? config["namespace"].as<std::string>() : name_;
          ^
                              != nullptr
```

marrts: > Does this not work?
> 
> ```
> ns_ = (YAML::Node n = config["namespace"]) ? n.as<std::string>() : name_;
> ```

That doesn't compile, but I could potentially do:

```C++
YAML::Node n;
ns_ = (n = config["namespace"]) ? config["namespace"].as<std::string>() : name_;
```

Levi-Armstrong: Okay, if CI passes then current state is good.

rjoomen:
Would you be able to add this to the documentation?

marrts:
> Would you be able to add this to the documentation?

I added [an example](https://github.com/tesseract-robotics/tesseract_planning/tree/master/tesseract_task_composer#reusing-a-namespace-across-multiple-tasks) in the README of `tesseract_task_composer`. Is there somewhere else you had in mind?

PR #443, opened 2024-02-23
Update docker.yml to allow workflow_dispatch

Levi-Armstrong:


marip8:
If you intend for this change to be able to push new docker images to the registry on command, then I think you will also have to update the `PUSH_DOCKER_IMAGE` variable. Right now it will only push when the reference branch is `master` (which may or may not be the case when you manually trigger this pipeline) or on `release` events. I looked into it briefly at one point and I wasn't able to quickly track down an event name associated with manual triggers

Levi-Armstrong:
@marip8 It looks like the docker file uses the dependencies.repos for both master and release docker creation. How can we update this so when building the master image it using the dependencies_unstable.repos?

Levi-Armstrong:
@marip8 or @marrts Do you either have time to look into how best to update the docker build to the unstable docker build pulls trajopt master?

marip8:
I can look into it

Levi-Armstrong:
This is no longer needed.

PR #447, opened 2024-03-18
Correctly handle cartesian waypoints in SimplePlannerFixedSizeAssignPlanProfile

rjoomen:
Previously, when the only second position was cartesian, the SimplePlannerFixedSizeAssignPlanProfile would take the joint position of the first position as seed, and when both were cartesian it would take the environment state as seed, leading strange planning results.
With this PR the 'seed' of the second cartesian position is used if available, or the IK of the second position is calculated using either the first position or the environment state as seed.

### DIFF
--- a/tesseract_motion_planners/simple/src/profile/simple_planner_fixed_size_assign_plan_profile.cpp
+++ b/tesseract_motion_planners/simple/src/profile/simple_planner_fixed_size_assign_plan_profile.cpp
@@ -24,6 +24,7 @@
  * limitations under the License.
  */
 
+#include <tesseract_command_language/cartesian_waypoint.h>
 #include <tesseract_motion_planners/simple/profile/simple_planner_fixed_size_assign_plan_profile.h>
 #include <tesseract_motion_planners/simple/interpolation.h>
 #include <tesseract_motion_planners/core/utils.h>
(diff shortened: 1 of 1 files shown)


Levi-Armstrong:
I will review today

Levi-Armstrong:
I do not believe this one is intended to run inverse kinematics.

Levi-Armstrong:
I do believe it should be updated to leverage the seed but should not run inverse kinematics. If I remember correctly this was intended to just used the current state.

rjoomen:
> I do believe it should be updated to leverage the seed but should not run inverse kinematics. If I remember correctly this was intended to just used the current state.

OK, but shouldn't it be called `SimplePlannerFixedSizeAssignNoIKPlanProfile` instead of `SimplePlannerFixedSizeAssignPlanProfile`, for clarity and to match `SimplePlannerLVSNoIKPlanProfile`?

Levi-Armstrong:
> > I do believe it should be updated to leverage the seed but should not run inverse kinematics. If I remember correctly this was intended to just used the current state.
> 
> OK, but shouldn't it be called `SimplePlannerFixedSizeAssignNoIKPlanProfile` instead of `SimplePlannerFixedSizeAssignPlanProfile`, for clarity and to match `SimplePlannerLVSNoIKPlanProfile`?

I am good with changing the name to make it more clear.

rjoomen:
I've done more work, including the NoIK renaming, and added it all to #448. Closing this PR.

PR #467, opened 2024-07-11
Add CI for Ubuntu Noble

rjoomen:


rjoomen:
This now fails because tesseract_planning has not been updated to match this yet: https://github.com/tesseract-robotics/tesseract/commit/0de0151cc99792e5836024fd6d56832503630a57

Levi-Armstrong:
This depends on the trajopt [MR](https://github.com/tesseract-robotics/trajopt/pull/410) so the docker is available.

rjoomen:
`noble-unstable` fails, because:
```
E: The repository 'https://ppa.launchpadcontent.net/levi-armstrong/tesseract-robotics/ubuntu noble Release' does not have a Release file.
```
This ppa is not necessary on Noble though, as `Taskflow` (3.6.0) and `qt-advanced-docking-system` (3.8.3) are available on Noble.

@Levi-Armstrong can you add a release file for Noble to the ppa (without packages), so the error does not occur, or do we need to modify the CI to not use the ppa at all for Noble? And if so, how?

Levi-Armstrong:
They are needed and I will work on getting them published.

Levi-Armstrong:
Oh, I missed that. If they are available on the system then you should be able to use those.

rjoomen:
> Oh, I missed that. If they are available on the system then you should be able to use those.

Indeed, I do. So the point is that the CI should not fail on your ppa not having an entry for Noble. It does not need to have the packages themselves, only a release file or something. Or if that is not possible, we'd need to disable the ppa for Noble.

rjoomen:
> Indeed, I do. So the point is that the CI should not fail on your ppa not having an entry for Noble. It does not need to have the packages themselves, only a release file or something. Or if that is not possible, we'd need to disable the ppa for Noble.

I have to correct myself: that won't work, as Taskflow is `taskflow` in your ppa, but `libtaskflow-cpp-dev` in Noble, and `qt-advanced-docking-system` vs `libqt-advanced-docking-system-dev`. How can we modify `.github/workflows/unstable.yml` to support both focal+jammy with the ppa and noble without, but with different package names?

Levi-Armstrong:
Replace the this with the following. I may not have the `if` syntax correct but should be possible.
```
- name: Install Depends
        shell: bash
        run: |
          apt update
          apt upgrade -y
          apt install -y wget software-properties-common
          add-apt-repository ppa:levi-armstrong/tesseract-robotics
          apt install -y libompl-dev taskflow
```

```
- name: Install Depends
        if: matrix.distro != 'noble'
        shell: bash
        run: |
          apt update
          apt upgrade -y
          apt install -y wget software-properties-common
          add-apt-repository ppa:levi-armstrong/tesseract-robotics
          apt install -y libompl-dev taskflow

- name: Install Depends
        if: matrix.distro == 'noble'
        shell: bash
        run: |
          apt update
          apt upgrade -y
          apt install -y libompl-dev libtaskflow-cpp-dev
```

rjoomen:
I have no idea why `rosdep` fails to detect the installation of  `google-perftools` in `Docker/noble`. It works fine for me locally in Noble.

Levi-Armstrong:
@rjoomen Are you planning to fix the clang-tidy errors; if not I recommend leaving the CI for clang-tidy to the one that was passing and create a ticket to updated clang-tidy.

rjoomen:
> @rjoomen Are you planning to fix the clang-tidy errors; if not I recommend leaving the CI for clang-tidy to the one that was passing and create a ticket to updated clang-tidy.

That's probably a better idea. I reverted the change.

Levi-Armstrong (approved):

PR #469, opened 2024-07-12
Remove TaskComposerProblem and leverage TaskComposerDataStorage instead

Levi-Armstrong:
The problem made the pipeline very ridged and confusing sense some data come from it and the TaskComposerDataStorage. This removed the TaskComposerProblem and moves everything into the data storage object. This make Task Composer pipelines very flexible now where you can have two different branches running each with a different environment modified profiles and much more.  

I also took some inspiration for BehaviorTreeCpp and introduced the concept of ports for the input and output keys. This way a key or keys is mapped by a port name to make it easier to reference in the code instead of an index. This will also make it easier when I get around to creating something similar to the BehaviorTreeCpp GUI. It also allows validation of the ports on construction to catch extra entries or missing required entries.


rjoomen:
Sounds very cool! I hope you will be able to include some examples and good documentation.

Levi-Armstrong:
> Sounds very cool! I hope you will be able to include some examples and good documentation.

I just updated the [readme](https://github.com/Levi-Armstrong/tesseract_planning/tree/feature/remove-task-composer-problem/tesseract_task_composer) so you can see difference. Now for each task you specify what it needs like environment, profiles, manip_info, etc.

marrts:
@Levi-Armstrong What is the plan for a new version tesseract being iterated with this change since it is a pretty big breaking change?

Levi-Armstrong:
I should be create a new release this week 0.24.0.

PR #489, opened 2024-07-27
Fix Windows and Mac OS Github Actions

johnwason:


johnwason:
Windows will not automatically export `const static` fields in classes such as these:

https://github.com/johnwason/tesseract_planning/blob/0c28710612a19baa1f2cb6e322d767e56cb39463/tesseract_task_composer/planning/include/tesseract_task_composer/planning/nodes/continuous_contact_check_task.h#L48

Either we need to manually export these symbols or use something other than `const static` fields.

Levi-Armstrong:
> Windows will not automatically export `const static` fields in classes such as these:
> 
> https://github.com/johnwason/tesseract_planning/blob/0c28710612a19baa1f2cb6e322d767e56cb39463/tesseract_task_composer/planning/include/tesseract_task_composer/planning/nodes/continuous_contact_check_task.h#L48
> 
> Either we need to manually export these symbols or use something other than `const static` fields.

I will add the export header and export the static variables.

johnwason:
It looks like the Mac OS build is going to need more upstream fixes. Let's merge this for now to get the Windows fixes.

Levi-Armstrong:
Why do you have to generate export file for each library? The files are identical except the name of the variable. Could we not just generate one for the core library and use it for everything?

.github/workflows/windows.yml
```diff
@@ -60,3 +60,4 @@ jobs:
         upstream-args: --cmake-args -G "Ninja" -DVCPKG_TARGET_TRIPLET=x64-windows-release -DCMAKE_BUILD_TYPE=Release -DBUILD_IPOPT=OFF -DBUILD_SNOPT=OFF -DTESSERACT_BUILD_TRAJOPT_IFOPT=OFF -DVCPKG_APPLOCAL_DEPS=OFF
         target-path: target_ws/src
         target-args: --packages-ignore tesseract_examples --cmake-args -G "Ninja" -DVCPKG_TARGET_TRIPLET=x64-windows-release -DCMAKE_BUILD_TYPE=Release -DTESSERACT_ENABLE_TESTING=ON -DTESSERACT_BUILD_TRAJOPT_IFOPT=OFF -DVCPKG_APPLOCAL_DEPS=OFF
+        run-tests-args: --packages-ignore tesseract_examples
```
Levi-Armstrong, line 63: What is the current issue with the examples?

johnwason: I haven't cleaned up the examples for Windows. I don't want to spend the time right now to work on them.

johnwason:
> Why do you have to generate export file for each library? The files are identical except the name of the variable. Could we not just generate one for the core library and use it for everything?

Each individual DLL needs to have its own import/export since they may reference each other.

johnwason:
I have the Mac OS working but I need to merge this pull request and a few others in different places.

Levi-Armstrong (approved):

PR #491, opened 2024-07-28
Fix windows build by exporting public static const class variables

Levi-Armstrong:


Levi-Armstrong:
@johnwason I believe this should fix the windows build issue in your open MR.

johnwason:
The export is applied to the class, and each individual library needs their own export macro. 

Here is an example of an export macro on a class:

https://github.com/robotraconteur/robotraconteur/blob/b0c1ccab189b059f3b59874a303a6a6f52e321d7/RobotRaconteurCore/include/RobotRaconteur/DataTypes.h#L250

I can take a look this week.

Levi-Armstrong:
I thought everything would get exported when using cmake `WINDOWS_EXPORT_ALL_SYMBOLS` and only these need to be exported manually.

johnwason:
> The module definition file will be passed to the linker causing all symbols to be exported from the .dll. For global data symbols, __declspec(dllimport) must still be used when compiling against the code in the .dll. All other function symbols will be automatically exported and imported by callers.

https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html

Data symbols still need to be exported. For classes with static fields I think the entire class needs to be exported but I haven't used this before.

PR #524, opened 2024-10-26
Fix use of trajopt composite profile cost and constraint collision co…

Levi-Armstrong:
…nfig


rjoomen:
Great find, but a pretty serious bug. Now I need to rethink the rule of thumb of `buffer ~ 2 * margin`.

PR #540, opened 2024-12-07
New ompl profile interface 

Levi-Armstrong:
The goal is to simplify the profiles interface and allow moving them to be loaded via plugin. As going through this I was able to main all of the same functionality along with the customization for the default profile. The initial thought was that we could go to a single profile for each planner but now I do not think that will be possible. We may be able to combine the solver and composite profile but we will have to have a distinction between the plan profile and composite profile. I do not currently see any way around this but open to ideas.


schornakj:
@Levi-Armstrong does this PR resolve [this TODO](https://github.com/tesseract-robotics/tesseract_planning/blob/82512cdcf1b12d4164ede4f0a482e9e49bd5ba95/tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp#L413-L417) from the version of the OMPL planner tagged in 0.27.0? This diff does change that part of the planner and it removes the TODO but I don't think that's specifically mentioned anywhere.

Levi-Armstrong:
It does.

PR #543, opened 2024-12-10
Print a collision summary to status of failed contact check

marrts:
Better debugging at a glance for dot graphs. Depends on https://github.com/tesseract-robotics/tesseract/pull/1079
Screenshot from 2024-12-10 11-30-00

### DIFF
--- a/tesseract_task_composer/planning/src/nodes/continuous_contact_check_task.cpp
+++ b/tesseract_task_composer/planning/src/nodes/continuous_contact_check_task.cpp
@@ -166,7 +166,18 @@ ContinuousContactCheckTask::runImpl(TaskComposerContext& context, OptionalTaskCo
   if (contactCheckProgram(contacts, *manager, *state_solver, ci, cur_composite_profile->config))
   {
     info->status_code = 0;
-    info->status_message = "Results are not contact free for process input: " + ci.getDescription();
+    std::stringstream ss;
+    ss << "Results are not contact free for process input: " << ci.getDescription() << "\n";
+    // Iterate over each contact and add collision summary if not empty
+    for (std::size_t i = 0; i < contacts.size(); ++i)
+    {
+      const auto& contact_map = contacts[i];
+      if (!contact_map.empty())
+      {
+        ss << contact_map.getCollisionSummary().str() << "\n";
+      }
+    }
+    info->status_message = ss.str();
     CONSOLE_BRIDGE_logInform("%s", info->status_message.c_str());
 
     // Save space
--- a/tesseract_task_composer/planning/src/nodes/discrete_contact_check_task.cpp
+++ b/tesseract_task_composer/planning/src/nodes/discrete_contact_check_task.cpp
@@ -162,7 +162,18 @@ std::unique_ptr<TaskComposerNodeInfo> DiscreteContactCheckTask::runImpl(TaskComp
   std::vector<tesseract_collision::ContactResultMap> contacts;
   if (contactCheckProgram(contacts, *manager, *state_solver, ci, cur_composite_profile->config))
   {
-    info->status_message = "Results are not contact free for process input: " + ci.getDescription();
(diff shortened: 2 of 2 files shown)


Levi-Armstrong:
Not that I am against adding this, just wanted mention here. There is now a log file that can be generated and tesseract_qt has a viewer that allows you to inspect a lot of this type of information.

marrts:
> There is now a log file that can be generated and tesseract_qt has a viewer that allows you to inspect a lot of this type of information.

I was unaware of this, I can't seem to get it to work in RVIZ with tesseract_ros2. I was able to save and then load a log file (although it was 6.6 GB), and I see things about viewing trajectories and toolpaths in the widget, but nothing actually displays.

In general my proposed change just helps for seeing things at a glance. I am saving all my programs at each step with unique names and publishing them all to the trajectory viewer, so I can debug things nicely, but it can be hard to dig through them all to find the collision and there are so many terminal logs that I can't easily find the collision report.

marrts:
> and there are so many terminal logs that I can't easily find the collision report

I think something that would be really cool is to have a tesseract logger object that optionally prints stuff to the terminal, but also stores all the logs to a map based on UUIDs of tasks/graphs. Then all the terminal logs could be exported into unique files, making it easy to debug an individual task even when multiple threads are running at the same time

tesseract_task_composer/planning/src/nodes/continuous_contact_check_task.cpp
```diff
@@ -160,7 +160,18 @@ ContinuousContactCheckTask::runImpl(TaskComposerContext& context, OptionalTaskCo
...
+    ss << "Results are not contact free for process input: " << ci.getDescription() << "\n";
+    // Iterate over each contact and add collision summary if not empty
+    for (std::size_t i = 0; i < contacts.size(); ++i)
+    {
+      const auto& contact_map = contacts[i];
+      if (!contact_map.empty())
+      {
+        ss << contact_map.getSummary() << "\n";
+      }
+    }
```
Levi-Armstrong, line 173: I could see this blowing up. Should we just print the first contact summary?

marrts: The way I have it, it's up to the user how much is printed. If they have contact check set to FIRST then they'll only get one. If it is set to ALL then they'll potentially get a long novel printed out, which is maybe useful for them.

Levi-Armstrong: I think it may be better to compile the vector of maps into a single map and then print the summary to avoid a large message.

Levi-Armstrong:
@marrts I think this just needs a few changes and then this can be merged in.

marrts:
> > @marrts I think this just needs a few changes and then this can be merged in.
> 
> If we're able to get [tesseract-robotics/tesseract#1200](https://github.com/tesseract-robotics/tesseract/pull/1200) merged then we could potentially have better information in the printout with the new return object. Maybe we wait for that. (I understand that would also require a bigger update to this for contactCheckProgram similar to checkTrajectory in tesseract_environment)

Just playing around with some stuff and we can make a printout that looks something like this:
image

image

Obviously, this can be changed based on feedback, but this makes it really easy to quickly see where and what happened.

Levi-Armstrong:
@marrts Do you have time to update this based on your changes in the tesseract repository?

marrts:
> @marrts Do you have time to update this based on your changes in the tesseract repository?

I should be able to do it later today or tomorrow.

marrts:
> > @marrts Do you have time to update this based on your changes in the tesseract repository?
> 
> I should be able to do it later today or tomorrow.

Pushed, but it does require this other tesseract PR to print out a condensed trajectory summary message. https://github.com/tesseract-robotics/tesseract/pull/1205

PR #553, opened 2024-12-18
Fixed clang-format in command language test suite

marip8:
Changes per commit message


Levi-Armstrong:
Yea, this is my fault. It seem that clang-format-12 on 20.04 produces different results than the same version on 22.04.

PR #555, opened 2024-12-23
Draft: Revise OMPL planner and profile to support different types of state spaces

marip8:
The current implementation of the OMPL planner effectively limits its usage to only the real vector state space (RVSS) because of the `OMPLStateExtractor` class and because of the way the planner converts OMPL paths into composite instructions. It would be beneficial to support other types of state spaces (specifically SE(3) space and the SE(3) space with constraints), and this is possible with some changes to the new profile interface:

## Removal of the `OMPLStateExtractor` concept

All of the existing components that utilize the state extractor are effectively only valid for use with RVSS. Removing the state extractor meants:
- Adding an RVSS-specific utility for converting an OMPL state to an `Eigen::VectorXd`
- Adding an RVSS-specific utility for converting an OMPL path to a `tesseract_common::TrajArray`
- Replacing calls to the state extractor with calls to these RVSS-specific utilities
- (TODO) renaming components (e.g., the state and motion validators) to indicate they are RVSS-specific components

## Conversion of OMPL paths into composite instructions

The motion planner itself currently attempts to convert OMPL paths to composite instructions by using the state extractor to turn the path into a joint trajectory, then converting the joint trajectory into a composite instruction of state or joint waypoints. To prevent this from limiting users to only the RVSS, an OMPL profile should provide a function to convert an OMPL path into a composite instruction (given an environment and manipulator info). This PR adds the following method to the base profile:

```c++
virtual CompositeInstruction convertPath(const ompl::geometric::PathGeometric& path,
                                         const tesseract_common::ManipulatorInfo& composite_mi,
                                         const std::shared_ptr<const tesseract_environment::Environment>& env) const = 0;
```

## To-do

The OMPL planner currently "assigns" the resulting OMPL paths into the output composite instruction via a process that also makes the assumption that the output should be state or joint waypoints (which would not be true for SE(3) state spaces that would utilize Cartesian waypoints). If the OMPL profile becomes responsible for converting an OMPL path into a composite instruction, then this strategy won't work.

The implementation in this PR simply concatenates all of the per-segment result composite instructions into a single composite instruction. I expect that this is probably not quite sufficient, so we'll need to think about how to handle this more effectively.


tesseract_motion_planners/ompl/include/tesseract_motion_planners/ompl/profile/ompl_profile.h
```diff
@@ -92,6 +85,13 @@ class OMPLPlanProfile : public Profile
                     const tesseract_common::ManipulatorInfo& composite_mi,
                     const std::shared_ptr<const tesseract_environment::Environment>& env) const = 0;
 
+  /**
+   * @brief Convert an OMPL defined path into a composite instruction that can be returned by the planner
+   */
+  virtual CompositeInstruction convertPath(const ompl::geometric::PathGeometric& path,
+                                           const tesseract_common::ManipulatorInfo& composite_mi,
+                                           const std::shared_ptr<const tesseract_environment::Environment>& env) const = 0;
```
marip8, line 93: To mirror what is currently being done in [`assignTrajectory`](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp#L205), do we also need to pass in the UUIDs of the start and end waypoints or those waypoints themselves?

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -60,43 +59,6 @@ using CachedSimpleSetupsPtr = std::shared_ptr<CachedSimpleSetups>;
...
-      if (extractor(goal_states->getState(i)).isApprox(state, 1e-5))
-        return true;
-  }
-  else
-  {
-    CONSOLE_BRIDGE_logWarn("checkGoalStates: Unsupported Goal Type!");
-    return true;
-  }
-  return false;
-}
```
marip8, line 98: Note, these functions were removed because they are RVSS-specific and don't really add any value at run-time IMO. If we're concerned that OMPL is not setting valid start and end states, we should either make a unit test to prove that this is not the case, or bake these checks in as real checks (not asserts) into the RVSS profile

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -293,14 +192,9 @@ PlannerResponse OMPLMotionPlanner::solve(const PlannerRequest& request) const
     cached_simple_setups.reserve(move_instructions.size());
 
   // Transform plan instructions into ompl problem
-  unsigned num_output_states = 1;
-  long start_index{ 0 };
-  std::size_t segment{ 1 };
```
marip8, line 298: Note: segment is the same as the iterator variable `i`, so no need to make something separate that needs to be incremented separately

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -318,27 +212,25 @@ PlannerResponse OMPLMotionPlanner::solve(const PlannerRequest& request) const
...
     }
     else
     {
-      simple_setup = cached_simple_setups.at(segment - 1);
+      simple_setup = cached_simple_setups.at(i - 1);
     }
 
     // Parallel Plan problem
-    auto status = parallelPlan(*simple_setup, *solver_config, num_output_states);
+    auto status = parallelPlan(*simple_setup, *solver_config, 2);
```
marip8, line 233: Same comment about `num_output_states`. It is initialized to 1, incremented by 1 and reset to 1 in every loop, making its value 2 every time it gets to this point.

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -348,33 +240,10 @@ PlannerResponse OMPLMotionPlanner::solve(const PlannerRequest& request) const
...
-                                   joint_names,
-                                   traj,
-                                   request.format_result_as_input);
-
-    // Reset data for next segment
-    start_instruction = end_instruction;
-    num_output_states = 1;
-    ++segment;
+    // Concatenate into results
+    std::copy(response.results.end(), output.begin(), output.end());
```
marip8, line 246: This is where I need a little help understanding what exactly `assignTrajectory` was doing so we can replicate that behavior correctly

Levi-Armstrong:
I think it is important that all motion planners preserve as much of the original program state. The program may include other instructions that need to be preserved. If it needs to insert more interpolation instructions that is fine but it should not ever reduce.

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
Levi-Armstrong: I think these should only be removed if a unit test is written in its place. I typically add assets for two reasons, the first if something is not easy to write a unit test and the second is I have not had time to write a unit test yet. They still have value in CI.

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
Levi-Armstrong: It is not the same as 'i'. The 'i' increments for every move instruction in a segment. A segment is incremented when a constrained waypoint is reached but can have any number of unconstrained waypoints in-between.

tesseract_motion_planners/ompl/include/tesseract_motion_planners/ompl/profile/ompl_profile.h
Levi-Armstrong: I  like this change. I was using the minimum needed but you could also pass the waypoint or instruction.

tesseract_motion_planners/ompl/src/continuous_motion_validator.cpp
```diff
@@ -139,8 +138,8 @@ bool ContinuousMotionValidator::continuousCollisionCheck(const ompl::base::State
   }
   mutex_.unlock();
 
-  Eigen::Map<Eigen::VectorXd> start_joints = extractor_(s1);
-  Eigen::Map<Eigen::VectorXd> finish_joints = extractor_(s2);
+  Eigen::Map<Eigen::VectorXd> start_joints = fromRealVectorStateSpace(s1, si_->getStateDimension());
+  Eigen::Map<Eigen::VectorXd> finish_joints = fromRealVectorStateSpace(s2, si_->getStateDimension());
```
Levi-Armstrong, line 142: How would this work with a constrained state space? The intent behind the extract was to avoid multiple implementations of this class.

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
Levi-Armstrong: The intent to is to preserve as much of the structure between two constrained waypoints. There can be any number of interpolation waypoints that should be preserved.

Levi-Armstrong:
Overall, I like the direction this heading.

tesseract_motion_planners/ompl/src/continuous_motion_validator.cpp
marip8: Check out 74d7dda, 8cbdbce, 27f376e, and 792626f, which add a constrained planner profile that inherits from the existing RVSS plan profile with minimal additional code. It at least compiles on my machine, but I haven't tested it yet.

I think we still need the concept of an `OMPLStateExtractor`, but it only applies to real vector state spaces, not all state spaces in general.

marip8:
> It would be beneficial to support other types of state spaces (specifically SE(3) space and the SE(3) space with constraint projection), and this is possible with some changes to the new profile interface:

Note: per the [constrained planning documentation](https://ompl.kavrakilab.org/constrainedPlanning.html#autotoc_md37), the constrained state space requires that the state be contiguous in memory, which means that we cannot directly use the OMPL SE(3) state space with constraints. We would have to make our own real vector SE(3) space (e.g., position and angle-axis) to support constrained planning in Cartesian space.

Levi-Armstrong:
@marip8 I will need to do a release this week. Are you wanting to get this into the next release?

PR #562, opened 2025-01-03
Fix ompl so it only simplifies when requested

Levi-Armstrong:
@marrts

### DIFF
--- a/tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
+++ b/tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
@@ -181,21 +181,9 @@ std::pair<bool, std::string> parallelPlan(ompl::geometric::SimpleSetup& simple_s
   if (solver_config.simplify)
     simple_setup.simplifySolution();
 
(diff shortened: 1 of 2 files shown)


tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
```diff
@@ -181,21 +181,9 @@ std::pair<bool, std::string> parallelPlan(ompl::geometric::SimpleSetup& simple_s
   if (solver_config.simplify)
     simple_setup.simplifySolution();
 
-  // Interpolate the path if it shouldn't be simplified and there are currently fewer states than requested
+  // Interpolate the path there are currently fewer states than requested
```
marrts, line 184: Minor typo. Should be "Interpolate the path **if** there are..."

marrts (commented):
Looks good to me other than the minor typo

tesseract_motion_planners/ompl/src/ompl_motion_planner.cpp
Levi-Armstrong: Updated

PR #574, opened 2025-01-06
Allow setting minimum timestamp delta between trajectory points in Iterative Spline Parameterization

schornakj:
I ran into a situation where I needed to enforce that the differences between timestamps for consecutive trajectory points were all greater than some minimum threshold. Briefly, this is because some robot controllers can't execute trajectories with arbitrarily-fine time resolution and need to have (for example) a 1ms difference between each trajectory point.

A good way to resolve this while ensuring that the trajectory is still smooth is to apply this time delta threshold during the time parameterization stage. This lets us stretch out the trajectory between two waypoints if needed while ensuring that the robot's path for this trajectory segment follows the same positions but just at a slower speed.

I'm using ISP specifically so I only implemented this change there. I did add this parameter to the time parameterizer base class to allow implementing it elsewhere as well. Let me know if there's a better way to approach this, though.

I have to test this more thoroughly on my system and then update the branch to the latest state of `main`, so I'll keep this PR as a draft until I do both those things.

### DIFF
--- a/tesseract_task_composer/planning/include/tesseract_task_composer/planning/profiles/iterative_spline_parameterization_profile.h
+++ b/tesseract_task_composer/planning/include/tesseract_task_composer/planning/profiles/iterative_spline_parameterization_profile.h
@@ -50,6 +50,8 @@ struct IterativeSplineParameterizationProfile
 
   /** @brief max_velocity_scaling_factor The max acceleration scaling factor passed to the solver */
   double max_acceleration_scaling_factor{ 1.0 };
+
+  double minimum_time_delta{ std::numeric_limits<double>::epsilon() };
 };
 }  // namespace tesseract_planning
 
--- a/tesseract_task_composer/planning/src/nodes/iterative_spline_parameterization_task.cpp
+++ b/tesseract_task_composer/planning/src/nodes/iterative_spline_parameterization_task.cpp
@@ -191,6 +191,8 @@ IterativeSplineParameterizationTask::runImpl(TaskComposerContext& context,
                                                  cur_composite_profile->max_acceleration_scaling_factor;
   Eigen::VectorXd jerk_scaling_factors = Eigen::VectorXd::Ones(static_cast<Eigen::Index>(flattened.size()));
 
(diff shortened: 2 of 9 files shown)


rjoomen:
Looking at MoveIt, I noticed that the TOTG algorithm nowadays supports spacing waypoints equally in time. As the TOTG algorithm of Tesseract [is forked from MoveIt](https://github.com/tesseract-robotics/tesseract_planning/pull/23), I think the best solution would be to bring Tesseract TOTG up to date with the [MoveIt](https://github.com/moveit/moveit/blob/master/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h)(1)/[MoveIt2](https://github.com/moveit/moveit2/blob/main/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.hpp) version. This was on my wishlist anyway, as this would bring in some improvements/fixes as well (see [here](https://github.com/moveit/moveit/commits/master/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp) and [here](https://github.com/moveit/moveit2/commits/main/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp)), so I'm willing to help.

Also, TOTG is proven to be superior to ISP, hence MoveIt2 dropped ISP, another reason for switching.

gavanderhoorn:
@schornakj: is this related to Yaskawa-Global/motoros2#255?

Levi-Armstrong:
@marip8 What are your thoughts on this?

schornakj:
@Levi-Armstrong @marip8 alrighty, this is ready for review.

I've been running this branch on my robot (a Yaskawa GP180-120 with a YRC1000 controller) and it resolves the issue I described in my comment in https://github.com/Yaskawa-Global/motoros2/issues/255. I added some tests to show that the new parameter does have an effect on the trajectory timestamps too.

MotoROS2 cannot execute trajectories with time deltas less than 1ms, so a change like this one is required to prevent the time parameterizer from generating a trajectory that the robot cannot execute.

PR #577, opened 2025-01-09
Include task composer in .run-cpack, fix typos in CMakeLists

MaxDeSantis:
- Adds task composer to `.run-cpack` array so debs get built
- Fixes dependency on `tesseract-descartes-core` which should be `descartes-light-core`
- Other typos

Note that I changed the motion planner cpack dependency `trajopt-ifopt` to `trajopt_ifopt`. Not sure if it was intended to be hyphenated or underscored, so I went with the less substantial fix. I believe Debian standard is all hyphens.


tesseract_motion_planners/descartes/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ if(TESSERACT_PACKAGE)
     DESCRIPTION "Tesseract motion planners descartes components"
     COMPONENT_DEPENDS core simple
     LINUX_DEPENDS "${TESSERACT_PACKAGE_PREFIX}tesseract-collision-core"
-                  "${TESSERACT_PACKAGE_PREFIX}tesseract-descartes-core"
+                  "${TESSERACT_PACKAGE_PREFIX}descartes-light-core"
     WINDOWS_DEPENDS "${TESSERACT_PACKAGE_PREFIX}tesseract-collision-core"
-                    "${TESSERACT_PACKAGE_PREFIX}tesseract-descartes-core")
+                    "${TESSERACT_PACKAGE_PREFIX}descartes-light-core")
```
Levi-Armstrong, line 62: @marip8 You good with this change. I think this should be good but I think you had added the tesseract prefix.

marip8: Yes, this was probably a typo

tesseract_task_composer/planning/CMakeLists.txt
```diff
@@ -94,8 +94,8 @@ endif()
 if(tesseract_motion_planners_trajopt_ifopt_FOUND)
   list(APPEND FACTORIES_SOURCE_FILES src/factories/planning_task_composer_trajopt_ifopt_plugin_factory.cpp)
   list(APPEND FACTORIES_SOURCE_LINK_LIBRARIES tesseract::tesseract_motion_planners_trajopt_ifopt)
-  list(APPEND LINUX_DEPENDS "${TESSERACT_PACKAGE_PREFIX}tesseract-motion-planners-trajopt-ifopt")
-  list(APPEND WINDOWS_DEPENDS "${TESSERACT_PACKAGE_PREFIX}tesseract-motion-planners-trajopt-ifopt")
+  list(APPEND LINUX_DEPENDS "${TESSERACT_PACKAGE_PREFIX}tesseract-motion-planners-trajopt_ifopt")
+  list(APPEND WINDOWS_DEPENDS "${TESSERACT_PACKAGE_PREFIX}tesseract-motion-planners-trajopt_ifopt")
```
marip8, line 98: I think this change is correct for the current version of RICB, but I think we should modify the `cpack_component` macro to also replace underscores in component names, which it currently does not. This seems like a non-standard name

marip8: See [here](https://github.com/ros-industrial/ros_industrial_cmake_boilerplate/pull/99) for the proposed fix

PR #580, opened 2025-01-12
Remove manip_info as option task input

Levi-Armstrong:


Levi-Armstrong:
This had already been removed from tesseract_ros planning request.

PR #596, opened 2025-02-21
Allow enabling OSQP workspace update via settings

rjoomen:
Accompanies https://github.com/tesseract-robotics/trajopt/pull/446


Levi-Armstrong (approved):

rjoomen:
I have a small update left enabling warm start if not updating. I'll push that later today. So far I have not seen any benefit for my application of either updating (with or without warm start) or explicit warm starting using previous results when not updating.

PR #599, opened 2025-03-05
Simplify the use of poly types

Levi-Armstrong:


marip8 (approved):
I like the way this looks now. I think this will make the creation of planning instructions cleaner and easier to understand

PR #605, opened 2025-04-02
Changes with legacy trajopt leveraging trajopt_common::TrajOptCollisionConfig

Levi-Armstrong:


marip8 (approved):
Looks reasonable to me once we figure out the details on the update to `trajopt`

Levi-Armstrong:
@marip8 Can you review the latest changes and make sure I did not miss anything?

tesseract_motion_planners/descartes/include/tesseract_motion_planners/descartes/impl/descartes_collision_edge_evaluator.hpp
```diff
@@ -45,21 +45,26 @@ template <typename FloatType>
...
   , active_link_names_(manip_->getActiveLinkNames())
   , discrete_contact_manager_(collision_env.getDiscreteContactManager())
   , continuous_contact_manager_(collision_env.getContinuousContactManager())
-  , collision_check_config_(std::move(config))
+  , collision_check_config_(std::move(collision_check_config))
   , allow_collision_(allow_collision)
   , debug_(debug)
 {
+  collision_check_config_.contact_request.type =
+      (allow_collision_) ? tesseract_collision::ContactTestType::CLOSEST : tesseract_collision::ContactTestType::FIRST;
```
marip8, line 61:
```suggestion
```

I think this should be removed and the user should be allowed to set the contact request type however they see fit

Levi-Armstrong: I am not sure about this one because this can greatly impact performance with no added benefit if they select anything other than these two. Should I just check the value and throw an exception if not configured correctly?

Levi-Armstrong: Lets open an issue to further discuss how best to handle this. I agree that its odd that it is be configured outside of the users control.

Issue #601, opened 2025-03-17
Avoid using `assert` for validation during runtime


schornakj:
In general I think it's a good idea to avoid using `assert` for things like enforcing that a joint trajectory satisfies the kinematic chain's joint limit.

[Here's an example in the Descartes planner](https://github.com/tesseract-robotics/tesseract_planning/blob/189269eb9b38b22be7b52ea1bd422f432dec0a29/tesseract_motion_planners/descartes/include/tesseract_motion_planners/descartes/impl/descartes_motion_planner.hpp#L160) -- it's possible to hit this assert even if other plans have succeeded during the same session if the program is composing joint waypoints at runtime.

Instead, these cases should create an error message describing what happened and return a failure status. This gives an opportunity for higher-level control logic to respond to the failure automatically if appropriate.

Happy to  write up some PRs for this myself, just wanted to post an issue about it before I got started.

Levi-Armstrong:
I think we made this change in some of the other planners which had this check.

Levi-Armstrong:
Addressed by #606

PR #606, opened 2025-04-10
Swap descartes assert with warning message

Levi-Armstrong:
Fix issue #601


Levi-Armstrong:
@schornakj you good with this change?

schornakj:
@Levi-Armstrong I think it's still correct to treat this as an error state and exit, since clamping the joint states to be within limits would change the FK solution and move the tip link relative to the base frame. Returning a PlannerResponse with `successful == false` would be the right thing to do IMO.

(sorry I wasn't able to jump on this myself right away -- thanks for starting to address it!)

PR #607, opened 2025-04-16
Fix TrajOptIfoptOSQPSolverProfile constructors

rjoomen:
The default move and move assign constructors of TrajOptIfoptOSQPSolverProfile caused a "double free or corruption (!prev)" crash on destruction of the class if it had been moved, for instance through serialization. This PR fixes this.

I also moved opt_params to the TrajOptIfoptSolverProfile class to match the structure of TrajOptSolverProfile/TrajOpOSQPSolverProfile.


rjoomen:
Forget this. This does use after move, which obviously is wrong.

PR #612, opened 2025-04-16
Add task composer task return value override

marip8:
This PR provides a means to override the return value of a task. This can be useful for debugging (e.g., forcing a task composer graph to take a specific path) and potentially for unit testing (for the same reason).

### DIFF
--- a/tesseract_task_composer/core/src/task_composer_node.cpp
+++ b/tesseract_task_composer/core/src/task_composer_node.cpp
@@ -131,7 +131,7 @@ int TaskComposerNode::run(TaskComposerContext& context, OptionalTaskComposerExec
     {
       results.return_value = return_value_override_;
       results.status_message += " (Return value overridden to " + std::to_string(return_value_override_) + ")";
-      results.color = "yellow";
+      results.color = "cyan";
     }
   }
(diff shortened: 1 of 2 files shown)


Levi-Armstrong (changes requested):
The only request is to add unit tests to cover the new functionality.

tesseract_task_composer/core/src/task_composer_node.cpp
```diff
@@ -158,6 +163,12 @@ int TaskComposerNode::run(TaskComposerContext& context, OptionalTaskComposerExec
   try
   {
     results = runImpl(context, executor);
+    if (return_value_override_ >= 0)
+    {
+      results.return_value = return_value_override_;
+      results.status_message += " (Return value overridden to " + std::to_string(return_value_override_) + ")";
+      results.color = "yellow";
```
marrts, line 170: Personally, I would vote for a color other than yellow as that is already used elsewhere. Maybe a blue or cyan? I was going to say gray, but that appears to already be used to represent "abort"

Levi-Armstrong:
@marip8 I assume we still want to get this merged in?

PR #620, opened 2025-05-09
Fix createCartesianPositionConstraint not using coeffs

rjoomen:
Fix forgotten coeffs param when creating CartPosConstraint.


rjoomen:
Thanks. I cannot merge this, as the Windows checks are mandatory and only Owners can override this, it seems.

PR #626, opened 2025-05-28
Task composer factory update

marip8:
This PR updates the task composer plugin factory class to:
- provide access to the names of all the known plugins.
- default-construct task and executor plugins
- load task and executor plugins when only the name is provided (i.e., no configuration information).
    - Previously, the plugin factory class would return an error if no configuration information had been given to the factory. With this change, the plugin gets default-constructed if no plugin information is provided.

### DIFF
--- a/tesseract_task_composer/core/src/task_composer_plugin_factory.cpp
+++ b/tesseract_task_composer/core/src/task_composer_plugin_factory.cpp
@@ -434,4 +434,15 @@ YAML::Node TaskComposerPluginFactory::getConfig() const
 
   return config;
 }
+
+std::vector<std::string> TaskComposerPluginFactory::getAvailableTaskComposerNodePlugins() const
+{
+  return impl_->plugin_loader.getAvailablePlugins("TaskNode");
+}
+
+std::vector<std::string> TaskComposerPluginFactory::getAvailableTaskComposerExecutorPlugins() const
+{
+  return impl_->plugin_loader.getAvailablePlugins("TaskExec");
+}
+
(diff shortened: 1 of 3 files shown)


tesseract_task_composer/core/src/task_composer_plugin_factory.cpp
```diff
@@ -282,17 +282,42 @@ std::string TaskComposerPluginFactory::getDefaultTaskComposerNodePlugin() const
 std::unique_ptr<TaskComposerExecutor>
 TaskComposerPluginFactory::createTaskComposerExecutor(const std::string& name) const
 {
-  const auto& executor_plugin_info = impl_->executor_plugin_info;
-  auto cm_it = executor_plugin_info.plugins.find(name);
-  if (cm_it == executor_plugin_info.plugins.end())
+  try
```
Levi-Armstrong, line 285: My only concern with this change is that the plugins are not intended to be reconfigurable after being created which can lead to incorrect usage. Instead what if this method generated a meta package (which needs to be defined) that this function produces. Once option I explored in the past was leveraging a yaml file which include this information which could also be used for the yaml editor in tesseract_qt.

Levi-Armstrong: Here is the schema I was exploring at the time. I think it would be good to discuss a little more because this could benefit all plugins throughout tesseract. 


https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_task_composer/config/task_composer_plugins.yaml.schema

PR #628, opened 2025-06-12
Removed redundant edge evaluator check for cartesian waypoint

marrts:
I'm not sure what the motivation behind this check for cartesian waypoints was, it does the exact same thing in and out of the if statement.

### DIFF
--- a/tesseract_motion_planners/descartes/include/tesseract_motion_planners/descartes/impl/profile/descartes_default_move_profile.hpp
+++ b/tesseract_motion_planners/descartes/include/tesseract_motion_planners/descartes/impl/profile/descartes_default_move_profile.hpp
@@ -150,20 +150,6 @@ std::unique_ptr<descartes_light::EdgeEvaluator<FloatType>> DescartesDefaultMoveP
 
   auto manip = DescartesMoveProfile<FloatType>::createKinematicGroup(manip_info, *env);
 
-  if (move_instruction.getWaypoint().isCartesianWaypoint())
-  {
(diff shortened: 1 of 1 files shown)


tesseract_motion_planners/descartes/include/tesseract_motion_planners/descartes/impl/profile/descartes_default_move_profile.hpp
```diff
@@ -150,20 +150,6 @@ std::unique_ptr<descartes_light::EdgeEvaluator<FloatType>> DescartesDefaultMoveP
...
-      return std::make_unique<descartes_light::EuclideanDistanceEdgeEvaluator<FloatType>>();
-
-    auto compound_evaluator = std::make_unique<descartes_light::CompoundEdgeEvaluator<FloatType>>();
-    compound_evaluator->evaluators.push_back(
-        std::make_shared<descartes_light::EuclideanDistanceEdgeEvaluator<FloatType>>());
-    compound_evaluator->evaluators.push_back(std::make_shared<DescartesCollisionEdgeEvaluator<FloatType>>(
-        *env, manip, edge_contact_manager_config, edge_collision_check_config, allow_collision, debug));
-
-    return compound_evaluator;
-  }
```
Levi-Armstrong, line 165: This was introduced when switching to the new profile to support plugins. I went back and verified that these two sections in the old version were the same to make sure a bug was not introduced. This is good to remove.

Levi-Armstrong (approved):

Issue #631, opened 2025-06-25
Task Composer Plugin Factory increasing constructor time on repeat calls


marrts:
### Version

master

### OS Version

Ubuntu 22.04

### Describe the bug

When constructing a `TaskComposerPluginFactory` with a re-used `YAML::Node`, each successive call to the constructor takes longer than the previous call. 

### To Reproduce

I added some tests to `tesseract_task_composer_plugin_factories_unit.cpp` on this branch on my fork: https://github.com/marrts/tesseract_planning/tree/bug/task_composer_plugin_factory_increasing_time

Look below for the output of the tests:

### Expected behavior

Each time calling the constructor should be consistent in the time it takes to construct the object. I think there is some weird YAML shared memory stuff going on.

### Relevant log output

```shell
[ RUN      ] TesseractTaskComposerFactoryUnit.LoadTaskComposerPluginFactoryMultiTimes
LoadFromYAMLNode - Iteration 0: 0.638938 seconds
LoadFromYAMLNode - Iteration 1: 1.01681 seconds
LoadFromYAMLNode - Iteration 2: 1.53459 seconds
LoadFromYAMLNode - Iteration 3: 2.00653 seconds
LoadFromYAMLNode - Iteration 4: 2.40333 seconds
LoadFromYAMLNode - Iteration 5: 2.90787 seconds
LoadFromYAMLNode - Iteration 6: 3.17379 seconds
LoadFromYAMLNode - Iteration 7: 3.65178 seconds
LoadFromYAMLNode - Iteration 8: 4.09297 seconds
LoadFromYAMLNode - Iteration 9: 4.53538 seconds
/tesseract_planning/tesseract_task_composer/test/tesseract_task_composer_plugin_factories_unit.cpp:334: Failure
Expected: (slope) < (mean * 0.10), actual: 0.431475 vs 0.25962
Detected a significant upward trend in timings (slope=0.43147530670302991) for test: LoadFromYAMLNode
[  FAILED  ] TesseractTaskComposerFactoryUnit.LoadTaskComposerPluginFactoryMultiTimes (25978 ms)
[ RUN      ] TesseractTaskComposerFactoryUnit.LoadTaskComposerPluginFactoryMultiTimesInLambda
LoadFromYAMLNodeInLambda - Iteration 0: 0.642663 seconds
LoadFromYAMLNodeInLambda - Iteration 1: 0.642031 seconds
[...]

PR #632, opened 2025-06-25
Fix YAML issue with TaskComposerPluginFactory constructor

marrts:
See #631


Levi-Armstrong:
This has been addressed by the following, [PR1](https://github.com/tesseract-robotics/tesseract/pull/1172) and [PR2](https://github.com/tesseract-robotics/tesseract_planning/pull/639)

PR #636, opened 2025-07-01
Correctly update input key rather than output key for raster task

marrts:
I think the UpdateEndStateTask, UpdateStartStateTask, and UpdateStartAndEndStateTask tasks were all being incorrectly created in the raster_motion_task and raster_only_mostion_task source files. Previously, the UpdateXTasks would output the output key of the freespace tasks, meaning their changes would not actually get applied to the input of the upcoming subgraph. This issue is masked if all input/outputs use the same keys.

## Before:
Screenshot from 2025-07-01 12-58-14

## After:
Screenshot from 2025-07-01 13-18-00


Levi-Armstrong (approved):

Levi-Armstrong:
Good catch

PR #637, opened 2025-07-03
Add trajopt joint configs for fix state collision task

marrts:
Previously the joint cost/constraint coefficients were hardcoded and the tolerance was hardcoded for all joints as the jiggle_factor. Also the tolerance only applied to the constraint and couldn't be applied to the cost. Also cost and constraints were always on for the joint waypoint. There are use cases where users are okay with the robot joint state moving more significantly, but they just want to find a somewhat close state out of collision.


Levi-Armstrong:
@marrts It looks like clang-tidy and possible unit tests are not happy.

Levi-Armstrong (approved):

PR #655, opened 2025-08-06
Add trajopt collision coefficients to fix state collision profile

marrts:
When running a pipeline I had a very small collision distance that was barely violated and I tried to run fix state collision with trajopt, but it would immediately converge despite still being in collision. It would make a negligible improvement, causing trajopt to stop and then it evaluated and considered the constraints sufficiently met. This meant that my subsequent collision check still failed, failing the pipeline. Exposing the coefficients to the collision constraint and cost let me set the collision constraint coefficient to 1000 and the same task consistently succeeded given the same motion plan.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        distro: [focal, jammy, noble]
+        distro: [jammy, noble]
     env:
       REGISTRY: ghcr.io
       IMAGE_NAME: ${{ github.repository }}
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        distro: [focal, jammy, noble]
+        distro: [jammy, noble]
     container:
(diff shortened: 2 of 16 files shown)


tesseract_task_composer/planning/src/profiles/fix_state_collision_profile.cpp
```diff
@@ -45,6 +45,8 @@ FixStateCollisionProfile::FixStateCollisionProfile(Settings mode)
   collision_check_config.type = tesseract_collision::CollisionEvaluatorType::DISCRETE;
   trajopt_joint_constraint_config.coeff = Eigen::VectorXd::Constant(1, 1, 1);
   trajopt_joint_cost_config.coeff = Eigen::VectorXd::Constant(1, 1, 5);
+  collision_constraint_coeff = trajopt_common::CollisionCoeffData(1.0);
+  collision_cost_coeff = trajopt_common::CollisionCoeffData(20.0);
```
Levi-Armstrong, line 49: Also need to add these to the constructor below for parsing from yaml file.

marrts: Realized the yaml constructor was also missing the trajopt join configs. I also realized those did not have yaml encode/decode methods, so I have another PR to add that here: https://github.com/tesseract-robotics/trajopt/pull/469

PR #667, opened 2025-09-05
Update now that osqp_eigen supports osqp v1.0.0 params

Levi-Armstrong:
This depends on the following osqp_eigen [PR](https://github.com/robotology/osqp-eigen/pull/214).


rjoomen:
That's quick!

PR #671, opened 2025-09-25
Add missing ONE_PER_STEP check in contactCheckProgram

marrts:
There was no check for the exit_condition `ONE_PER_STEP` making it effectively the same as `ALL`


rjoomen:
LGTM

PR #672, opened 2025-09-25
Add solver settings to fix state collision task

marrts:
More control over trajopt solver settings for fix state collision. I ran into a problem where I had to change some settings for it to converge, so I thought it would make sense just to expose them. If you don't touch these settings, I believe everything should stay unchanged.


Levi-Armstrong:
@marrts Thanks for the improvement.

PR #680, opened 2025-11-14
Restore optimization callback functionality

rjoomen:


Levi-Armstrong:
I will review this week. Since the profiles need to have serialization we need to see how best to make it possible with this change. My thoughts originally was to provide the interface to retrieve the callbacks but not the member variable because it does not have serialization. Then the developer can just inherit and override the method to return the desired callbacks. The other option is to see how best to annd serialization to the callback. I have not looked at this in depth so let me know if you see any other options. Also I know we are close to switching profiles to be loaded as plugins which likely remove the need to be serialized.

Levi-Armstrong:
You should be good to rebase

Levi-Armstrong (approved):

PR #710, opened 2026-01-30
Always return motion planner response

rjoomen:
The motion planner result was not returned if the planner failed. So for example for a SimplePlanner-Trajopt pipeline, one would only receive the result of the SimplePlanner if Trajopt failed. The partially optimized but failed Trajopt result certainly is of interest for understanding why it failed, so it should be returned. The planner still properly responds with failure with this PR.


rjoomen:
@Levi-Armstrong CI failures are because of Trajopt updates, not this PR.

rjoomen:
Isn't error branching already supported through [this line](https://github.com/tesseract-robotics/tesseract_planning/blob/master/tesseract_task_composer/planning/include/tesseract_task_composer/planning/nodes/motion_planner_task.hpp#L188)?

PR #713, opened 2026-02-03
Remove fixed indices from optimization

rjoomen:
I noticed Trajopt Ifopt adding a JointPos constraint to the optimization for constrained/fixed joint waypoints. Legacy Trajopt does not do this. Is this intentional? If not, this PR has a fix (suggested by Copilot, to be fair). What do you think, @Levi-Armstrong?


rjoomen:
These todos in the code are still open, but apart from that I don't  think solving it is that relevant anymore.

Levi-Armstrong:
>  I suppose in Trajopt Ifopt I explicitly need to set costs and constraints for start and end, right?

That is correct with TrajOpt Ifopt you have to be explicit for the start and end state if it is a joint waypoint.

rjoomen:
Would checking if the number of (non-fixed) axes is smaller or equal than the number of fixed Cartesian components (x, y, z, rx, ry, rz) work? So for a fully fixed Cartesian pose a 7-axis robot is (possibly, excluding singularities and joint limits) redundant, but one with less axes not, whereas for a pose with, say, rx free a 6-axis robot is redundant.  Not sure if this always holds though.

PR #715, opened 2026-02-04
Allow setting minimum delta_t between traj points in ISP solver

Levi-Armstrong:
Replaces PR #574


Levi-Armstrong:
@schornakj Thanks for the improvement. Sorry for taking so long to get to this.

PR #718, opened 2026-02-04
Only make joint waypoint fixed if not tolerances and the constraint config enabled

Levi-Armstrong:


tesseract_motion_planners/trajopt_ifopt/src/profile/trajopt_ifopt_default_move_profile.cpp
```diff
@@ -234,8 +234,9 @@ TrajOptIfoptDefaultMoveProfile::create(const MoveInstructionPoly& move_instructi
       jwp = move_instruction.getWaypoint().as<JointWaypointPoly>();
       if (jwp.isConstrained())
       {
-        // Add to fixed indices
-        if (!jwp.isToleranced()) /** @todo Should not make fixed if term_type is cost */
+        // Add to fixed indices if constraint config is enabled and not tolernaced to prevent collision checks at this
```
rjoomen: tolernaced -> toleranced

PR #734, opened 2026-04-13
Move from string-based links and joints to id-based logic (using hashes)

rjoomen:
Requires https://github.com/tesseract-robotics/tesseract/pull/1274 and https://github.com/tesseract-robotics/trajopt/pull/554

For reference: this was squash merged, the full 86 commit history is backed up here: https://github.com/rjoomen/tesseract_planning/commits/archive/pr-734-pre-squash


command_language/src/utils.cpp
```diff
@@ -218,66 +291,95 @@ Eigen::VectorXd getJointPosition(const std::vector<std::string>& joint_names, co
...
-      jv = &(cwp.getSeed().position);
-      jn = &(cwp.getSeed().joint_names);
-    }
-    else
-    {
+    if (!cwp.hasSeed())
+      return false;
+
+    auto& seed = cwp.getSeed();
+    const std::vector<common::JointId>& seed_ids = seed.joint_ids;
```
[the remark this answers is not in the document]
rjoomen: Fixed by https://github.com/tesseract-robotics/tesseract_planning/pull/739 (this branch will be rebased on master once the PR is merged).

command_language/src/utils.cpp
```diff
@@ -218,66 +291,95 @@ Eigen::VectorXd getJointPosition(const std::vector<std::string>& joint_names, co
...
+
+      long idx = std::distance(seed_ids.begin(), it);
+      output(static_cast<long>(i)) = seed.position(idx);
     }
+
+    seed.joint_ids = joint_ids;
+    seed.position = output;
+
+    return true;
   }
```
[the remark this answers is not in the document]
rjoomen: Fixed by https://github.com/tesseract-robotics/tesseract_planning/pull/739 (this branch will be rebased on master once the PR is merged).

command_language/src/utils.cpp
```diff
@@ -218,66 +291,95 @@ Eigen::VectorXd getJointPosition(const std::vector<std::string>& joint_names, co
...
+    long idx = std::distance(jn_local.begin(), it);
     output(static_cast<long>(i)) = (*jv)(idx);
   }
 
-  *jn = joint_names;
+  if (waypoint.isJointWaypoint())
+    waypoint.as<JointWaypointPoly>().setJointIds(joint_ids);
+  else if (waypoint.isStateWaypoint())
+    waypoint.as<StateWaypointPoly>().setJointIds(joint_ids);
   *jv = output;
```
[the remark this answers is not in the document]
rjoomen: Fixed by https://github.com/tesseract-robotics/tesseract_planning/pull/739 (this branch will be rebased on master once the PR is merged).

command_language/test/command_language_utils_unit.cpp
```diff
@@ -314,13 +316,66 @@ TEST(TesseractCommandLanguageUtilsUnit, formatJointPositionTests)  // NOLINT
...
-  EXPECT_ANY_THROW(formatJointPosition({ "joint_1" }, wp0_poly));             // NOLINT
-  EXPECT_ANY_THROW(formatJointPosition({ "joint_3", "joint_1" }, wp0_poly));  // NOLINT
+  EXPECT_ANY_THROW(formatJointPosition(std::vector<std::string>{ "joint_1" }, wp0_poly));             // NOLINT
+  EXPECT_ANY_THROW(formatJointPosition(std::vector<std::string>{ "joint_3", "joint_1" }, wp0_poly));  // NOLINT
 
   WaypointPoly error_poly;
   EXPECT_ANY_THROW(formatJointPosition(format_joint_names, error_poly));  // NOLINT
 }
 
+TEST(TesseractCommandLanguageUtilsUnit, formatJointPositionByIdTests)  // NOLINT
```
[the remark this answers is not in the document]
rjoomen: Fixed by https://github.com/tesseract-robotics/tesseract_planning/pull/739 (this branch will be rebased on master once the PR is merged).

command_language/test/command_language_utils_unit.cpp
```diff
@@ -314,13 +316,66 @@ TEST(TesseractCommandLanguageUtilsUnit, formatJointPositionTests)  // NOLINT
...
+  EXPECT_EQ(wp00_poly.as<JointWaypointPoly>().getJointIds(), format_joint_ids);
+
+  WaypointPoly wp1_poly{ wp1 };
+  EXPECT_TRUE(formatJointPosition(format_joint_ids, wp1_poly));
+  EXPECT_TRUE(wp1_poly.as<CartesianWaypointPoly>().getSeed().position.isApprox(format_seed_position));
+  EXPECT_EQ(wp1_poly.as<CartesianWaypointPoly>().getSeed().joint_ids, format_joint_ids);
+
+  WaypointPoly wp1u_poly{ wp1 };
+  EXPECT_FALSE(formatJointPosition(joint_ids, wp1u_poly));
+  EXPECT_TRUE(wp1u_poly.as<CartesianWaypointPoly>().getSeed().position.isApprox(seed_position));
```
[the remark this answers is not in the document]
rjoomen: Fixed by https://github.com/tesseract-robotics/tesseract_planning/pull/739 (this branch will be rebased on master once the PR is merged).

command_language/src/utils.cpp
```diff
@@ -218,66 +291,95 @@ Eigen::VectorXd getJointPosition(const std::vector<std::string>& joint_names, co
...
-  if (jn->size() != joint_names.size())
-    throw std::runtime_error("Joint name sizes do not match!");
+  if (jn_local.size() != joint_ids.size())
+    throw std::runtime_error("Joint ID sizes do not match!");
 
-  if (joint_names == *jn)
+  if (joint_ids == jn_local)
     return false;
 
   Eigen::VectorXd output = *jv;
```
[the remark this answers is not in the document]
rjoomen: Fixed by https://github.com/tesseract-robotics/tesseract_planning/pull/739 (this branch will be rebased on master once the PR is merged).

command_language/src/utils.cpp
```diff
@@ -218,66 +291,95 @@ Eigen::VectorXd getJointPosition(const std::vector<std::string>& joint_names, co
...
+      auto it = std::find(seed_ids.begin(), seed_ids.end(), joint_ids[i]);
+      if (it == seed_ids.end())
+        throw std::runtime_error("Joint IDs do not match!");
+
+      long idx = std::distance(seed_ids.begin(), it);
+      output(static_cast<long>(i)) = seed.position(idx);
     }
+
+    seed.joint_ids = joint_ids;
+    seed.position = output;
```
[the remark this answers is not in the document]
rjoomen: Fixed by https://github.com/tesseract-robotics/tesseract_planning/pull/739 (this branch will be rebased on master once the PR is merged).

command_language/src/joint_waypoint.cpp
```diff
@@ -4,47 +4,49 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
...
+#include <tesseract/common/types.h>
 #include <tesseract/command_language/joint_waypoint.h>
 
 namespace tesseract::command_language
 {
-// NOLINTNEXTLINE(modernize-pass-by-value)
-JointWaypoint::JointWaypoint(std::vector<std::string> names, const Eigen::VectorXd& position, bool is_constrained)
-  : names_(std::move(names)), position_(position), is_constrained_(is_constrained)
+JointWaypoint::JointWaypoint(std::vector<tesseract::common::JointId> joint_ids,
+                             Eigen::VectorXd position,
```
Levi-Armstrong: Eigen type needs to be passed by const reference.

command_language/src/joint_waypoint.cpp
```diff
@@ -4,47 +4,49 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
...
-                             const Eigen::VectorXd& lower_tol,  // NOLINT(modernize-pass-by-value)
-                             const Eigen::VectorXd& upper_tol)  // NOLINT(modernize-pass-by-value)
-  : names_(std::move(names))
-  , position_(position)
-  , lower_tolerance_(lower_tol)
-  , upper_tolerance_(upper_tol)
+JointWaypoint::JointWaypoint(std::vector<tesseract::common::JointId> joint_ids,
+                             Eigen::VectorXd position,
+                             Eigen::VectorXd lower_tol,
+                             Eigen::VectorXd upper_tol)
```
Levi-Armstrong: Eigen type needs to be passed by const reference.

command_language/src/state_waypoint.cpp
```diff
@@ -33,37 +33,43 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
 }
-StateWaypoint::StateWaypoint(const std::vector<std::string>& names,
-                             const Eigen::VectorXd& position,      // NOLINT(modernize-pass-by-value)
-                             const Eigen::VectorXd& velocity,      // NOLINT(modernize-pass-by-value)
-                             const Eigen::VectorXd& acceleration,  // NOLINT(modernize-pass-by-value)
+
+StateWaypoint::StateWaypoint(std::vector<tesseract::common::JointId> joint_ids,
+                             Eigen::VectorXd position,
+                             Eigen::VectorXd velocity,
+                             Eigen::VectorXd acceleration,
```
Levi-Armstrong: Eigen type needs to be passed by const reference.

motion_planners/descartes/src/descartes_collision.cpp
```diff
@@ -36,32 +36,31 @@ DescartesCollision::DescartesCollision(const tesseract::environment::Environment
...
 
 bool DescartesCollision::validate(tesseract::collision::ContactResultMap& contact_results,
-                                  tesseract::common::TransformMap& transforms_cache,
+                                  tesseract::common::LinkIdTransformMap& transforms_cache,
                                   const Eigen::Ref<const Eigen::VectorXd>& pos)
 {
   // Happens in two phases:
   // 1. Compute the transform of all objects
   contact_results.clear();
-  transforms_cache.clear();
```
Levi-Armstrong, line 64: I think we need to leave this?

motion_planners/descartes/src/descartes_collision.cpp
```diff
@@ -72,13 +71,12 @@ bool DescartesCollision::validate(tesseract::collision::ContactResultMap& contac
...
 
 double DescartesCollision::distance(tesseract::collision::ContactResultMap& contact_results,
-                                    tesseract::common::TransformMap& transforms_cache,
+                                    tesseract::common::LinkIdTransformMap& transforms_cache,
                                     const Eigen::Ref<const Eigen::VectorXd>& pos)
 {
   // Happens in two phases:
   // 1. Compute the transform of all objects
   contact_results.clear();
-  transforms_cache.clear();
```
Levi-Armstrong, line 81: I think we need to keep this?

motion_planners/ompl/src/collision_cost_objective.cpp
```diff
@@ -74,12 +74,18 @@ ompl::base::Cost CollisionCostObjective::stateCost(const ompl::base::State* stat
   mutex_.unlock();
 
   Eigen::Map<Eigen::VectorXd> finish_joints = extractor_(state);
-  TESSERACT_THREAD_LOCAL tesseract::common::TransformMap state1;
-  state1.clear();
```
Levi-Armstrong, line 78: This is a thread local so I think it would be good to clear here.

motion_planners/ompl/src/continuous_motion_validator.cpp
```diff
@@ -143,16 +143,25 @@ bool ContinuousMotionValidator::continuousCollisionCheck(const ompl::base::State
   Eigen::Map<Eigen::VectorXd> start_joints = extractor_(s1);
   Eigen::Map<Eigen::VectorXd> finish_joints = extractor_(s2);
 
-  TESSERACT_THREAD_LOCAL tesseract::common::TransformMap state0;
-  state0.clear();
```
Levi-Armstrong, line 147: I think we should still clear

motion_planners/ompl/src/continuous_motion_validator.cpp
```diff
@@ -143,16 +143,25 @@ bool ContinuousMotionValidator::continuousCollisionCheck(const ompl::base::State
...
+  TESSERACT_THREAD_LOCAL tesseract::common::LinkIdTransformMap state0;
   manip_->calcFwdKin(state0, start_joints);
 
-  TESSERACT_THREAD_LOCAL tesseract::common::TransformMap state1;
-  state1.clear();
```
Levi-Armstrong, line 151: I think we should still clear.

motion_planners/ompl/src/state_collision_validator.cpp
```diff
@@ -74,12 +74,16 @@ bool StateCollisionValidator::isValid(const ompl::base::State* state) const
   mutex_.unlock();
 
   Eigen::Map<Eigen::VectorXd> finish_joints = extractor_(state);
-  TESSERACT_THREAD_LOCAL tesseract::common::TransformMap state1;
-  state1.clear();
```
Levi-Armstrong, line 78: I think the clear should remain.

motion_planners/simple/src/interpolation.cpp
```diff
@@ -100,8 +101,8 @@ JointGroupInstructionInfo::~JointGroupInstructionInfo() = default;
 
 Eigen::Isometry3d JointGroupInstructionInfo::calcCartesianPose(const Eigen::VectorXd& jp, bool in_world) const
 {
-  TESSERACT_THREAD_LOCAL tesseract::common::TransformMap transforms;
-  transforms.clear();
```
Levi-Armstrong, line 104: I think the clear is still needed.

motion_planners/simple/src/interpolation.cpp
```diff
@@ -171,8 +172,8 @@ KinematicGroupInstructionInfo::~KinematicGroupInstructionInfo() = default;
 
 Eigen::Isometry3d KinematicGroupInstructionInfo::calcCartesianPose(const Eigen::VectorXd& jp, bool in_world) const
 {
-  TESSERACT_THREAD_LOCAL tesseract::common::TransformMap transforms;
-  transforms.clear();
```
Levi-Armstrong, line 175: I think the clear is still needed

task_composer/planning/src/nodes/fix_state_collision_task.cpp
```diff
@@ -70,12 +70,13 @@ bool stateInCollision(const Eigen::Ref<const Eigen::VectorXd>& start_pos,
...
+  manager->setActiveCollisionObjects(joint_group->getActiveLinkIds());
   manager->applyContactManagerConfig(profile.contact_manager_config);
 
-  TESSERACT_THREAD_LOCAL tesseract::common::TransformMap state;
-  state.clear();
   contacts.clear();
+  // Reused without clearing; safe here — calcFwdKin overwrites this group's links (see its contract) and
+  // only those links are read below, so a stale key from a prior call is harmless.
+  TESSERACT_THREAD_LOCAL tesseract::common::LinkIdTransformMap state;
   joint_group->calcFwdKin(state, start_pos);
```
Levi-Armstrong, line 79: Thes state needs to be cleared first.

task_composer/planning/src/nodes/fix_state_collision_task.cpp
```diff
@@ -334,21 +335,19 @@ bool moveWaypointFromCollisionTrajopt(tesseract::command_language::WaypointPoly&
     TESSERACT_THREAD_LOCAL tesseract::collision::ContactResultMap collisions;
     collisions.clear();
 
-    TESSERACT_THREAD_LOCAL tesseract::common::TransformMap state;
-    state.clear();
-
-    pci.kin->calcFwdKin(state, start_pos);
```
Levi-Armstrong, line 340: Do you think the thread local is not needed anymore?

command_language/src/joint_waypoint.cpp
rjoomen: Agreed — this was a misguided clang-tidy `modernize-pass-by-value` fix. Restored in 7726b0e093.

command_language/src/joint_waypoint.cpp
rjoomen: Same misguided clang-tidy fix — restored in 7726b0e093.

command_language/src/state_waypoint.cpp
rjoomen: Same misguided clang-tidy fix — restored in 7726b0e093.

motion_planners/descartes/src/descartes_collision.cpp
rjoomen: `JointGroup::calcFwdKin` assigns a transform for every link in the scene graph on every call, unconditionally, so there is nothing stale for the clear to protect against.

This is the documented contract on `JointGroup::calcFwdKin` and on both `StateSolver::getLinkTransforms` overloads — they don't clear, and reuse is safe as long as every key later read is a link of the group. The code is in fact stronger than the doc, which scopes it to the group rather than the whole scene.

Following it through: `JointGroup`'s constructor always builds a `KDLStateSolver`, over a tree from `parseSceneGraph` whose DFS adds a segment for every scene-graph vertex — and the constructor asserts `static_link_ids_.size() + active_link_ids_set_.size() == scene_graph.getLinks().size()`. `calculateTransformsHelper` then walks that whole tree from the root doing `link_transforms[link_id] = global_frame` per segment, with no conditional. The trailing `transforms.insert(static_link_transforms_...)` is a strict no-op, since every key is already assigned.

Nothing outside that set is ever read here: `checkTrajectoryState` iterates `manager.getActiveCollisionObjects()`, which returns `active_` verbatim — set once from `manip_->getActiveLinkIds()`, and `applyContactManagerConfig` only touches margins, ACM and object-enabled. Active links are a subset of scene links, so every key read was just written. Same for the `distance` overload below.

motion_planners/descartes/src/descartes_collision.cpp
rjoomen: Same reasoning as `validate` above — see https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3807895763. `checkTrajectoryState` only reads `manager.getActiveCollisionObjects()`, all of which `calcFwdKin` just assigned.

motion_planners/ompl/src/collision_cost_objective.cpp
rjoomen: Mechanism in https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3807895763. Here the only reads are `state1.at(id)` for `id ∈ link_ids_`, and `link_ids_` is a copy of `manip_->getActiveLinkIds()` — all assigned by the `calcFwdKin` on the line above. This is the path that runs per state evaluation, so it's where the clear cost actually shows.

motion_planners/ompl/src/continuous_motion_validator.cpp
rjoomen: Mechanism in https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3807895763. `state0` is only read as `state0.at(id)` for `id ∈ link_ids_`, a copy of `manip_->getActiveLinkIds()`, every one of which `calcFwdKin` assigns.

motion_planners/ompl/src/continuous_motion_validator.cpp
rjoomen: Same as `state0` just above — see https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3807895763.

motion_planners/ompl/src/state_collision_validator.cpp
rjoomen: Mechanism in https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3807895763. Reads are confined to `state1.at(id)` over `link_ids_`, a copy of `manip_->getActiveLinkIds()`.

motion_planners/simple/src/interpolation.cpp
rjoomen: Mechanism in https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3807895763 — `tcp_frame` is a link of the same scene graph the group was built from, so `calcFwdKin` always assigns it.

These two were worth a second look though, for a reason adjacent to your comment: they read `transforms[tcp_frame]` rather than `.at()`, and `Isometry3d`'s default constructor leaves the matrix uninitialized. If `tcp_frame` were ever absent, `operator[]` inserts that garbage into the map — and without the clear it would persist rather than being rebuilt each call. `tcp_frame` is also the only `ManipulatorInfo` field with no existence check; `working_frame` throws at construction via `getLinkTransform`, and `tcp_offset` throws inside `findTCPOffset`. Switched both to `.at()` in f42bfaa43f.

motion_planners/simple/src/interpolation.cpp
rjoomen: Same as `JointGroupInstructionInfo::calcCartesianPose` above — see https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3807951361, including the `.at()` change.

task_composer/planning/src/nodes/fix_state_collision_task.cpp
rjoomen: Mechanism in https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3807895763. `checkTrajectoryState` reads only `manager.getActiveCollisionObjects()`, set two lines up to `joint_group->getActiveLinkIds()` — every one of which `calcFwdKin` just assigned.

task_composer/planning/src/nodes/fix_state_collision_task.cpp
rjoomen: Fair point — this runs per waypoint via `applyCorrectionWorkflow`, so on a long trajectory with several failed corrections it is entered repeatedly, and I shouldn't have treated it as a one-shot path.

The reason it still seemed worth dropping is that the same block calls `env->getDiscreteContactManager()`, which clones the cached manager on every entry — so a `LinkIdTransformMap` allocation alongside it isn't measurable. The value-returning `JointGroup::calcFwdKin` is also the safer of the two here, since `setCollisionObjectsTransform` takes the whole map and iterates it, setting every key it recognises; a reused map still carrying links from another environment would move objects that shouldn't move.

Happy to restore the thread-local if you'd rather keep it symmetric with the `ContactResultMap` above it.

rjoomen:
Added some doc comment fixes, and added a check to createNearJointStateTermInfo to make it consistent with the other TermInfos. All inspired by @Levi-Armstrong's review. Also increased test coverage to make CodeCov happy (patch 76.89% -> 83.29, project 80.26% -> 81.33%.

rjoomen:
Added a commit that completes #535, which introduced the "size 6 or 1, replicated" tolerance contract but implemented it only in createCartesianWaypointTermInfo. Now it's also in createDynamicCartesianWaypointTermInfo.

motion_planners/descartes/src/descartes_collision.cpp
Levi-Armstrong: I do not think this is an issue when using a single environment. However, if you set up a planning server that manages any number of environments using the same thread pool, these thread_local objects could grow indefinitely without calling clear(), since each environment would have a different set of links.

Calling `clear()` prevents this growth and ensures the capacity only needs to accommodate the environment with the largest number of links.

rjoomen: Good point, I'm looking into this (might be a wider issue, as I did more cleaning up of TLS and .clear()), I'll get back to you.

Levi-Armstrong:
@rjoomen Are you still looking into the thread_local/clear; I just want to make sure you are not waiting on me?

rjoomen:
> @rjoomen Are you still looking into the thread_local/clear; I just want to make sure you are not waiting on me?

No problem, and yes, will further look into it and reply tomorrow.

motion_planners/descartes/src/descartes_collision.cpp
rjoomen: You're right, and it's a case my analysis didn't cover — I argued correctness within one environment, where every key read is one `calcFwdKin` just assigned. Residency across environments sharing a thread pool is a separate question, and the map does end up holding the union of every link set the thread has served.

Restored in af6dad6c90: ten `clear()` calls across seven files, back to the upstream form. The `getRobotConfig` membership guard went with them — it existed only because `at()` could find a stale key, which can't happen once the map is cleared. One site keeps the change: the `moveWaypointFromCollisionTrajopt` failure path no longer holds a thread-local at all, just a local from the value-returning `calcFwdKin`, so there is nothing there to accumulate.

The clears did show up in profiling — `clear()` frees every node and keeps only the bucket array, so it costs one allocation per link per call — which is why they came out. Bounding residency is worth more than that. It's probably worth revisiting the design later so we can have both; moving the scratch into the object that owns it would do it.

While checking whether we had done this anywhere else, I found the same pattern already in trajopt master, in two commits with opposite outcomes. 184926b4 moved the `CollisionEvaluator` transform caches from `static thread_local` to instance members in the same change that dropped their clears, so those are bounded by the evaluator's lifetime — that is the shape worth aiming for. 43b3cbad dropped five clears in `kinematic_terms.cpp` while the caches stayed `TrajOptVectorOfVector::transforms_cache` and `TrajOptMatrixOfVector::transforms_cache`, both `static thread_local`, so those five do have the problem you describe. I'll raise that separately. tesseract itself is clean — its scratch maps all still clear.

motion_planners/descartes/src/descartes_collision.cpp
rjoomen: Superseded by https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3829772967 — you were right that a thread_local scratch map accumulates the union of every link set the thread has served once you have several environments on one thread pool. The clear is restored here in `af6dad6c90`.

motion_planners/ompl/src/collision_cost_objective.cpp
rjoomen: Superseded by https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3829772967 — you were right that a thread_local scratch map accumulates the union of every link set the thread has served once you have several environments on one thread pool. The clear is restored here in `af6dad6c90`.

motion_planners/ompl/src/continuous_motion_validator.cpp
rjoomen: Superseded by https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3829772967 — you were right that a thread_local scratch map accumulates the union of every link set the thread has served once you have several environments on one thread pool. The clear is restored here in `af6dad6c90`.

motion_planners/ompl/src/continuous_motion_validator.cpp
rjoomen: Superseded by https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3829772967 — you were right that a thread_local scratch map accumulates the union of every link set the thread has served once you have several environments on one thread pool. The clear is restored here in `af6dad6c90`.

motion_planners/ompl/src/state_collision_validator.cpp
rjoomen: Superseded by https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3829772967 — you were right that a thread_local scratch map accumulates the union of every link set the thread has served once you have several environments on one thread pool. The clear is restored here in `af6dad6c90`.

motion_planners/simple/src/interpolation.cpp
rjoomen: Superseded by https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3829772967 — you were right that a thread_local scratch map accumulates the union of every link set the thread has served once you have several environments on one thread pool. The clear is restored here in `af6dad6c90`.

The other half of that reply still stands: `transforms[tcp_frame]` is now `transforms.at(tcp_frame)` as of `f42bfaa43f`.

motion_planners/simple/src/interpolation.cpp
rjoomen: Superseded by https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3829772967 — you were right that a thread_local scratch map accumulates the union of every link set the thread has served once you have several environments on one thread pool. The clear is restored here in `af6dad6c90`.

The other half of that reply still stands: `transforms[tcp_frame]` is now `transforms.at(tcp_frame)` as of `f42bfaa43f`.

task_composer/planning/src/nodes/fix_state_collision_task.cpp
rjoomen: Superseded by https://github.com/tesseract-robotics/tesseract_planning/pull/734#discussion_r3829772967 — you were right that a thread_local scratch map accumulates the union of every link set the thread has served once you have several environments on one thread pool. The clear is restored here in `af6dad6c90`.

rjoomen:
> Are you still looking into the thread_local/clear; I just want to make sure you are not waiting on me?

@Levi-Armstrong Done.

Levi-Armstrong (approved):

motion_planners/descartes/src/descartes_collision.cpp
rjoomen: Raised: tesseract-robotics/trajopt#579. It restores the five clears, switches the 14 cache reads from `operator[]` to `at()`, and adds a regression test that evaluates a calculator against a 93-link environment and then a 7-link one on the same thread — it fails if the clear is removed again.

One correction to what I wrote above. I called 184926b4's instance-member shape the one worth aiming for, and then did not take it. Moving the cache into the calculators would bound residency and keep the allocation saving, but `static thread_local` stays correct if an instance is ever shared across threads and a `mutable` member does not, with nothing enforcing the invariant. So I restored the clears instead. The PR spells out the alternatives I tried; happy to switch to the member form if you prefer it.

PR #739, opened 2026-04-23
Fix formatJointPosition & FormatAsInputTask aux-vector reorder bugs

### LINKED ISSUE #738: Joint reordering utilities leave tolerance/velocity/acceleration/effort vectors in original order
Summary

Two related sites in tesseract_planning reorder a waypoint's joint identifiers and position vector but leave joint-correlated auxiliary vectors in the original order. After the call, those auxiliary vectors silently correspond to the wrong joints.

Site 1 — formatJointPosition

tesseract::command_language::formatJointPosition in command_language/src/utils.cpp reorders position (and any seed.position for Cartesian waypoints) to match a caller-supplied joint ordering, then overwrites the stored joint identifiers. It does not permute:

- JointWaypoint::lower_tolerance_, JointWaypoint::upper_tolerance_
- StateWaypoint → JointState::velocity, acceleration, effort
- CartesianWaypoint seed velocity / acceleration / effort

Both the pre-existing std::vector<std::string> overload on master and the new std::vector<JointId> overload introduced by PR #734 have this behavior — the ID overload faithfully preserved the older logic.

Site 2 — FormatAsInputTask

task_composer/planning/src/nodes/format_as_input_task.cpp:144-151:

auto& jwp = mi.getWaypoint().as<JointWaypointPoly>();
if (!jwp.isConstrained() || (jwp.isConstrained() && jwp.isToleranced()))
{
jwp.setJointIds(getJointIds(umi.getWaypoint()));
jwp.setPosition(getJointPosition(umi.getWaypoint()));
}

This path specifically targets toleranced JointWaypoints. It replaces both joint_ids_ and position_ with the post-planning ordering but never touches lower_tolerance_ / upper_tolerance_. If post-planning joint order differs from the user's input ordering — the normal case, since otherwise the reformatting step
would be a no-op — the tolerances silently mis-associate.

Impact

formatProgram is invoked from FormatPlanningInputTask (task_composer/planning/src/nodes/format_planning_input_task.cpp:111) on every planning request that uses the standard input-formatting pipeline, so Site 1 is on the hot path whenever that task is wired up. Site 2 affects the same planner hookups that use
FormatAsInputTask to push post-planning state back into the pre-planning program.

Downstream, trajopt_ifopt_utils.cpp:109-110 consumes the tolerances as:

lower_limit = position + getLowerTolerance();
upper_limit = position + getUpperTolerance();

A mis-ordered tolerance vector produces incorrect joint bounds for the optimizer. Symptoms:

- Silent: no exception, no assertion.
- Only triggers when the supplied ordering differs from the ordering already stored in the waypoint — which is exactly the reason these utilities get called.

Reproduction

Any program whose waypoint joint ordering differs from the manipulator's getJointNames() / getJointIds(), run through either formatProgram (Site 1) or FormatAsInputTask (Site 2). The tolerance-bound mismatch is hard to spot without instrumenting the trajopt constraints.

Proposed fix

For both sites:

1. Compute the permutation index map once (same approach already used for position in Site 1).
2. Apply the same permutation to every joint-correlated vector on the waypoint when its size matches position.
3. Handle mismatched/empty auxiliary vectors defensively — skip when the auxiliary vector is empty, throw when non-empty and size doesn't match position.
4. Add unit tests that exercise a non-identity permutation and verify all auxiliary vectors move with position. Include both JointWaypoint (tolerances) and StateWaypoint (velocity/acceleration/effort) cases, plus the FormatAsInputTask path.

rjoomen:
## Summary
- `formatJointPosition` now applies the same joint-index permutation to every non-empty joint-correlated vector (JointWaypoint tolerances, StateWaypoint velocity/acceleration/effort, Cartesian seed velocity/acceleration/effort). Previously only `position` and `seed.position` were reordered; the rest silently mis-associated.
- `FormatAsInputTask`'s toleranced-JointWaypoint branch now routes through `formatJointPosition` before overwriting `position`, so tolerances follow the new joint ordering.
- Regression tests cover all three waypoint types plus the `FormatAsInputTask` scenario.

## Context
Fixes #738.

Downstream impact: `trajopt_ifopt_utils.cpp` consumes tolerances as `position ± tolerance` to build joint bounds, so a mis-ordered tolerance vector produced incorrect bounds silently.

## Test plan
- [x] New unit tests cover JointWaypoint tolerances, StateWaypoint velocity/accel/effort (including empty-aux case), Cartesian seed aux vectors, and FormatAsInputTask end-to-end.
- [x] \`colcon test --packages-select tesseract_planning\` passes (284/284).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/command_language/src/utils.cpp
+++ b/command_language/src/utils.cpp
@@ -218,64 +218,96 @@ Eigen::VectorXd getJointPosition(const std::vector<std::string>& joint_names, co
   return output;
 }
 
+namespace
+{
+/// @brief Build an index permutation such that out[i] = in[perm[i]] gives `in` reordered to `target_names`.
+/// Throws if sizes mismatch or a target name is absent from `current_names`.
+std::vector<std::size_t> buildJointPermutation(const std::vector<std::string>& target_names,
+                                               const std::vector<std::string>& current_names)
+{
+  if (current_names.size() != target_names.size())
+    throw std::runtime_error("Joint name sizes do not match!");
+
+  std::vector<std::size_t> perm(target_names.size());
+  for (std::size_t i = 0; i < target_names.size(); ++i)
+  {
+    auto it = std::find(current_names.begin(), current_names.end(), target_names[i]);
+    if (it == current_names.end())
+      throw std::runtime_error("Joint names do not match!");
+    perm[i] = static_cast<std::size_t>(std::distance(current_names.begin(), it));
+  }
+  return perm;
+}
+
+/// @brief Apply `perm` to `vec` in-place. No-op when `vec` is empty.
+/// Throws when `vec` is non-empty but its size does not match `perm`.
+void applyJointPermutation(Eigen::VectorXd& vec, const std::vector<std::size_t>& perm)
+{
+  if (vec.size() == 0)
+    return;
+  if (static_cast<std::size_t>(vec.size()) != perm.size())
+    throw std::runtime_error("Joint-correlated vector size does not match joint count!");
+
+  Eigen::VectorXd out(vec.size());
+  for (std::size_t i = 0; i < perm.size(); ++i)
+    out(static_cast<Eigen::Index>(i)) = vec(static_cast<Eigen::Index>(perm[i]));
+  vec = std::move(out);
+}
+}  // namespace
+
 bool formatJointPosition(const std::vector<std::string>& joint_names, WaypointPoly& waypoint)
 {
-  Eigen::VectorXd* jv{ nullptr };
-  std::vector<std::string>* jn{ nullptr };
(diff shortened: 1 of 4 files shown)


rjoomen:
One thing still to look into: the other vectors that are now ordered as well are checked for size. It might be the case that they are size one to be replicated to the correct number of joint later. If that is the case, size()==1 should so be accepted.

Edit: this is also the one line CodeCov complains about. Will look into this and add a test.

Levi-Armstrong (approved):

Levi-Armstrong:
Looks like clang-tidy is not happy.

PR #744, opened 2026-05-04
Mac Build Fixes

johnwason:


johnwason:
@Levi-Armstrong this is ready

task_composer/core/include/tesseract/task_composer/task_composer_log.h
```diff
@@ -21,6 +21,7 @@
 #include <memory>
 #include <tesseract/common/fwd.h>
 #include <tesseract/task_composer/task_composer_data_storage.h>
+#include <tesseract/common/serialization_extensions.h>
```
Levi-Armstrong, line 24: Does this have to go in the header?

task_composer/core/src/task_composer_keys.cpp
```diff
@@ -79,6 +81,7 @@ bool TaskComposerKeys::operator!=(const TaskComposerKeys& rhs) const { return !o
 
 std::ostream& operator<<(std::ostream& os, const TaskComposerKeys& keys)
 {
+  using namespace std;
```
Levi-Armstrong, line 84: Instead of including the full namespace can we just use the namespace where it is needed?

Levi-Armstrong (changes requested):

task_composer/core/include/tesseract/task_composer/task_composer_log.h
johnwason: searialization_extensions.h is required for TESSERACT_CLASS_EXTENSION later in the header

task_composer/core/src/task_composer_keys.cpp
johnwason: The `using namespace std` is required on Mac for some of the output stream operators. They were not being resolved correctly.

Levi-Armstrong (approved):

PR #742, opened 2026-05-01
Windows Build Fixes

johnwason:
This PR fixes the Windows build.

PR #751, opened 2026-06-01
Restore Cereal template registration in compiled cpp files instead of…

rjoomen:
… via _impl headers for non-Windows builds.

See also https://github.com/tesseract-robotics/tesseract/pull/1305

PR https://github.com/tesseract-robotics/tesseract_planning/pull/742 caused a regression in build memory usage and build times. Reason was the heavy Cereal template instantiation whenever these headers were included. This PR reverts the template registration to cpp files for non-Windows builds.

This also adds [CEREAL_FORCE_DYNAMIC_INIT](https://uscilab.github.io/cereal/assets/doxygen/polymorphic_8hpp.html#a8e0d5df9830c0ed7c60451cf2f873ff5) in the header (non-Windows) to match CEREAL_REGISTER_DYNAMIC_INIT in the cpp.

Windows builds are not affected, the fix of the mentioned PR is still in place for that platform.

### DIFF
--- a/.github/workflows/clang_format.yml
+++ b/.github/workflows/clang_format.yml
@@ -22,5 +22,5 @@ jobs:
           ./.run-clang-format
           if [ $? -ge 1 ]; then return 1; fi
           output=$(git diff)
-          echo $output
+          echo "$output"
           if [ -n "$output" ]; then exit 1; else exit 0; fi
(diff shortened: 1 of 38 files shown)


Levi-Armstrong:
@rjoomen I restarted CI now that the trajopt docker has been updated.

Levi-Armstrong (approved):

Levi-Armstrong:
@rjoomen It looks like conda-mac is failing here. I assume it needs the same fix you did in the tesseract repo.

rjoomen:
> @rjoomen It looks like conda-mac is failing here. I assume it needs the same fix you did in the tesseract repo.

That fix is already in, it's a different cause. I'm investigating.

rjoomen:
Fixed

PR #756, opened 2026-07-07
Fix CartesianWaypoint::print() to include rotation

NickNair:
## Summary

`CartesianWaypoint::print()` only printed the XYZ translation and had a `// TODO: Add rotation` comment. This PR completes it by printing the rotation as a quaternion (xyzw).

## Change

```cpp
const Eigen::Quaterniond q(transform_.rotation());
std::cout << prefix << "Cart WP: xyz=..." << " xyzw=" << q.x() << ", " << q.y() << ", " << q.z() << ", " << q.w() << std::endl;
```
### Test

Verified with basic_cartesian_example in debug mode output now shows:
```
Cart WP: xyz=0.5, -0.2, 0.62 xyzw=0, 1, 0, 0
```
Note: This is a trivial fix but a deliberate first contribution to get familiar with the codebase and contribution workflow.

### DIFF
--- a/command_language/src/cartesian_waypoint.cpp
+++ b/command_language/src/cartesian_waypoint.cpp
@@ -27,10 +27,10 @@ void CartesianWaypoint::print(const std::string& prefix) const
 {
   const Eigen::Quaterniond q(transform_.rotation());
   std::cout << prefix << "Cart WP: xyz=" << transform_.translation().x() << ", " << transform_.translation().y()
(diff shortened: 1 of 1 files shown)


Levi-Armstrong (approved):

NickNair:
The required `Unstable/ noble-unstable (pull_request)` check fail seems to come from `.github/workflows/unstable.yml` as per the co-pilot explanation.

PR #757, opened 2026-07-08
Fix ContinuousMotionValidator reporting invalid lastValid after first collision

NickNair:
## Summary

OMPL relies on `lastValid` to identify the furthest reachable state along a motion before the first constraint violation. In `ContinuousMotionValidator::checkMotion`, the loop over path segments sets `is_valid = false` on collision but continues iterating subsequent invalid segments overwrite `lastValid.second` with a position beyond the first collision boundary, causing the planner to treat an infeasible state as the last reachable point and potentially extend the tree into collision.

The fix adds `break` after the first detected violation in both the state validity and continuous collision branches. This aligns the continuous checker with `DiscreteMotionValidator`, which already handles this correctly with an identical `break`.

## Test plan
- [x] Built from source and ran `freespace_ompl_example` 
- [x] `lastValid` position is now always at or before the first collision boundary

### DIFF
--- a/motion_planners/ompl/src/continuous_motion_validator.cpp
+++ b/motion_planners/ompl/src/continuous_motion_validator.cpp
@@ -87,6 +87,7 @@ bool ContinuousMotionValidator::checkMotion(const ompl::base::State* s1,
           state_space.interpolate(s1, s2, lastValid.second, lastValid.first);
 
         is_valid = false;
+        break;
       }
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
@NickNair Is this ready to review? I noticed a few things are not checked off in the PR Message.

NickNair:
> @NickNair Is this ready to review? I noticed a few things are not checked off in the PR Message.

Yes they are already addressed via the unit test

PR #758, opened 2026-07-09
Fix joint limit enforcement discarded in OMPL JointWaypoint start/goal registration

NickNair:
## Summary

- `applyGoalStates` and `applyStartStates` (JointGroup overloads) compute a limit-enforced copy of the waypoint (`solution`) via `satisfiesLimits`/`enforceLimits`, but then populate the OMPL state from the original `joint_waypoint` vector thereby discarding the clamped values.
- A joint value marginally outside the state space bounds (within the `1e-6` satisfiesLimits tolerance) would pass through unclamped into OMPL, producing an out-of-bounds start or goal state.
- The IK overloads of both functions already index `solution[j]` correctly  this aligns the JointWaypoint overloads with that established pattern.


Levi-Armstrong (approved):
Good catch

PR #763, opened 2026-07-17
Use the Cartesian waypoint seed when solving IK in the OMPL move profile

### LINKED ISSUE #762: OMPL profile re-solves IK for Cartesian waypoints from a hardcoded zero seed, ignoring the waypoint's attached seed state
## Summary

`OMPLRealVectorMoveProfile` solves start/goal states for `CartesianWaypoint`s by calling `calcInvKin` with a **zero joint seed**, even when the waypoint carries a seed state (`CartesianWaypointPoly::getSeed()` populated — e.g. by `MinLengthTask`'s interpolation). With seed-dependent kinematics solvers that return at most one solution (e.g. the KDL-LMA plugin), the zero-seed IK solution regularly lands outside joint limits. `calcInvKin`'s joint-limit filter then empties the goal set and the planner hard-aborts with:

```
All goal states are either in collision or outside limits
```

The practical consequence: programs whose LINEAR `StateWaypoint`s were converted to seeded `CartesianWaypoint`s by `MinLengthTask` fail **deterministically**, even though every original state is reachable and within joint limits. The seed that would make the IK succeed is sitting right on the waypoint — it just isn't used.

## Versions / environment

- tesseract_planning 0.34.0 (`d7dfa425ca`)
- tesseract 0.34.1 (`e3ee23cd`)
- tesseract_ros2 0.34.0
- Ubuntu 24.04 / ROS 2 Jazzy, GCC 13
- Kinematics: KDL-LMA inverse kinematics plugin (seed-dependent, single-solution)
- Application: a 6-DOF industrial arm

Relevant code: `tesseract_motion_planners_ompl/src/profile/ompl_real_vector_move_profile.cpp`, around line 298 (the `applyGoalStates` / `applyStartStates` region), where `calcInvKin` is invoked with a zero-initialized seed vector.

## Deterministic reproduction

1. Build a composite program containing LINEAR moves defined as `StateWaypoint`s (all states verified reachable and within joint limits).
2. Run it through a task-composer pipeline that executes `MinLengthTask` before the OMPL planner. `MinLengthTask` interpolates and emits `CartesianWaypoint`s with their seed states populated (`getSeed()` returns the interpolated joint state).
3. Configure a KDL-LMA kinematics plugin for the manipulator (any seed-dependent, ≤1-solution IK solver reproduces this; analytic multi-solution solvers can mask it).
4. Plan.

Result: `calcInvKin` is called with a zero seed for the goal waypoint; LMA converges to a solution outside the joint limits; the limit filter discards it; the goal-state set is empty; the planner throws `All goal states are either in collision or outside limits`.

In our application this reproduces byte-identically both offline (a replay harness feeding the recorded program straight into the pipeline) and live. We additionally confirmed at the disassembly level of the built `.so` that the waypoint's seed member is never consulted on this path — the IK seed is unconditionally the zero vector.

## Analysis — the seed is ignored

- `MinLengthTask` (and other upstream stages) deliberately attach a joint seed to the Cartesian waypoints they produce; `CartesianWaypointPoly::getSeed()` exists precisely so downstream consumers can use it.
- `OMPLRealVectorMoveProfile::applyGoalStates` / `applyStartStates` ignore it and pass a hardcoded zero seed to `calcInvKin`.
- For redundant or seed-sensitive solvers this is not just suboptimal: with LMA the single returned solution is a function of the seed, so a zero seed can (and in practice does) yield the out-of-limits branch of the solution manifold, turning a fully feasible program into a guaranteed failure.
- Because the failure depends only on the geometry and the (fixed) zero seed, it is deterministic — retries and restarts of the planner do not help.

## Suggested fix

In `applyGoalStates` / `applyStartStates`:

1. **Honor the attached seed**: if `wp.hasSeed()` / `getSeed()` is populated, use that joint state as the `calcInvKin` seed instead of the zero vector.
2. **Multi-restart fallback**: when no seed is present (or the seeded solve yields no in-limit solution), retry `calcInvKin` from multiple random in-limit seeds rather than a single zero seed. This also rescues robots without analytic IK plugins, where a single deterministic seed gives poor coverage of the solution space.

We validated a local patch implementing exactly this (use `getSeed()` when present in `applyGoalStates`/`applyStartStates`, random-restart fallback otherwise) against 0.34: the previously failing programs plan successfully, with no regressions observed in our other pipelines. Happy to turn it into a PR if the approach is acceptable.

munich35:
## Summary
`OMPLRealVectorMoveProfile` now uses the seed state attached to a `CartesianWaypoint` when solving IK for start/goal states, instead of always seeding `calcInvKin` with a hardcoded zero vector.

## Root cause
In `applyStartStates` / `applyGoalStates` (the `KinGroupIKInput` overloads), `calcInvKin` was called with `Eigen::VectorXd::Zero(dof)` even when the waypoint carried a seed (`CartesianWaypointPoly::getSeed()` populated, e.g. by `MinLengthTask`'s interpolation). For seed-dependent solvers that return a single solution (e.g. the KDL-LMA plugin), the zero-seed solution can converge to an out-of-limits branch of the solution manifold. `calcInvKin`'s joint-limit filter then discards it, the goal-state set ends up empty, and the planner aborts with:

```
All goal states are either in collision or outside limits
```

The failure is deterministic: programs whose LINEAR `StateWaypoint`s were converted to seeded `CartesianWaypoint`s by `MinLengthTask` fail even though every state is reachable and within limits — the seed that would make IK succeed is on the waypoint but was never consulted.

## Fix
- Thread the waypoint's seed into the two `KinGroupIKInput` `applyStartStates`/`applyGoalStates` overloads via an optional trailing `const Eigen::VectorXd& seed` parameter (defaulted, so the public API and existing zero-seed behavior are unchanged for callers that pass nothing).
- At the call sites, pass `cur_wp.getSeed().position` when `cur_wp.hasSeed()`.
- Inside, the seed is only used when its size matches the manipulator DOF; otherwise it falls back to the previous zero seed.

This also resolves the pre-existing `@todo Need to also provide the seed instruction to use here` in `applyStartStates`.

## Notes for review
- **Seed ordering**: `getSeed().position` is ordered by `getSeed().joint_names`. This follows the same convention already relied on elsewhere for Cartesian-waypoint seeds (e.g. `simple_motion_planner.cpp` and the simple-planner assign profiles use `getSeed().position` directly), so no reordering is applied here. The size guard makes a mismatched seed a safe no-op (falls back to zero); a seed is a hint, so a mis-ordered one degrades gracefully rather than breaking correctness.
- **Scope**: this implements the "honor the attached seed" part only. A multi-random-restart fallback for the no-seed / no-analytic-IK case (mentioned in the issue) would be a reasonable follow-up but is intentionally left out to keep this change minimal.

## Verification
An equivalent seed-honoring patch was validated in our downstream application on 0.34 (the previously-failing `MinLengthTask` -> OMPL -> KDL-LMA programs planned successfully, with no regressions observed elsewhere); this PR ports that approach to current `master`. The seed-selection logic and the call-site expressions were compile-checked in isolation against Eigen. A full `colcon` build against current `master` (0.35.0) was **not** run in this environment. Happy to add a unit test if you'd like.

Closes #762

### DIFF
--- a/motion_planners/ompl/include/tesseract/motion_planners/ompl/profile/ompl_real_vector_move_profile.h
+++ b/motion_planners/ompl/include/tesseract/motion_planners/ompl/profile/ompl_real_vector_move_profile.h
@@ -98,12 +98,14 @@ protected:
   static void applyGoalStates(ompl::geometric::SimpleSetup& simple_setup,
                               const tesseract::kinematics::KinGroupIKInput& ik_input,
                               const tesseract::kinematics::KinematicGroup& manip,
-                              tesseract::collision::DiscreteContactManager& contact_checker);
+                              tesseract::collision::DiscreteContactManager& contact_checker,
+                              const Eigen::VectorXd& seed = Eigen::VectorXd());
 
   static void applyStartStates(ompl::geometric::SimpleSetup& simple_setup,
                                const tesseract::kinematics::KinGroupIKInput& ik_input,
                                const tesseract::kinematics::KinematicGroup& manip,
-                               tesseract::collision::DiscreteContactManager& contact_checker);
+                               tesseract::collision::DiscreteContactManager& contact_checker,
+                               const Eigen::VectorXd& seed = Eigen::VectorXd());
 
   static void applyGoalStates(ompl::geometric::SimpleSetup& simple_setup,
                               const Eigen::VectorXd& joint_waypoint,
--- a/motion_planners/ompl/src/profile/ompl_real_vector_move_profile.cpp
+++ b/motion_planners/ompl/src/profile/ompl_real_vector_move_profile.cpp
@@ -245,22 +245,31 @@ void OMPLRealVectorMoveProfile::applyGoalStates(ompl::geometric::SimpleSetup& si
   const auto dof = manip.numJoints();
   tesseract::common::KinematicLimits limits = manip.getLimits();
 
-  /** @brief Making this thread_local does not help because it is not called enough during planning */
-  tesseract::kinematics::IKSolutions joint_solutions;
+  auto goal_states = std::make_shared<ompl::base::GoalStates>(simple_setup.getSpaceInformation());
+  std::vector<tesseract::collision::ContactResultMap> contact_map_vec;
+
   // Use the seed attached to the Cartesian waypoint when one is provided, otherwise fall back to the
   // zero seed. For seed-dependent solvers that return a single solution (e.g. KDL-LMA), a hardcoded
   // zero seed can converge to an out-of-limits branch of the solution manifold; the limit filter then
   // discards it and an otherwise-reachable goal ends up with an empty solution set.
-  Eigen::VectorXd ik_seed = Eigen::VectorXd::Zero(dof);
-  if (seed.size() == static_cast<Eigen::Index>(dof))
-    ik_seed = seed;
-  manip.calcInvKin(joint_solutions, { ik_input }, ik_seed);
-  auto goal_states = std::make_shared<ompl::base::GoalStates>(simple_setup.getSpaceInformation());
-  std::vector<tesseract::collision::ContactResultMap> contact_map_vec(static_cast<std::size_t>(joint_solutions.size()));
+  //
+  // The waypoint seed is used PREFERENTIALLY, not exclusively: a seed produced upstream without
+  // collision awareness (e.g. the simple planner's interpolation states) can equally steer a
+  // single-solution solver onto a branch that is in collision. If the seeded pass contributes no
+  // valid goal state, a second pass with the legacy zero seed recovers the previous behavior so a
+  // reachable goal is never lost to an unlucky seed.
+  const bool has_wp_seed = (seed.size() == static_cast<Eigen::Index>(dof));
 
-  for (std::size_t i = 0; i < joint_solutions.size(); ++i)
-  {
-    Eigen::VectorXd& solution = joint_solutions[i];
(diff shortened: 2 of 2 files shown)


Levi-Armstrong:
@munich35 It looks like the OMPL unit tests are failing which is interesting. Let me know what you find.

munich35:
Found it — the failure was real and caused by the change, thanks for flagging it.

**Root cause:** `generateInterpolatedProgram` attaches the simple planner's final interpolation state as the Cartesian waypoint's seed — and that seed is produced without collision awareness. In `OMPLFreespaceCartesianGoalPlannerUnit` the seeded IK (single-solution solver) converges onto a branch where `link_4` is ~8 mm inside the test box (`Solution: 0 Links: box_attached, link_4 Distance: -0.008`); the lone solution gets rejected, `GoalStates` ends up empty, and `createSetup` throws before any planner runs — which is why all eight configurators failed identically.

**Fix pushed (a812033):** the waypoint seed is now used *preferentially, not exclusively* — the seeded IK pass runs first, and a second pass with the legacy zero seed recovers the previous behavior whenever the seeded pass contributes no valid state. Applied symmetrically to `applyGoalStates` and `applyStartStates`, so a reachable goal/start is never lost to an unlucky seed while the original issue (#762 — a zero seed converging out-of-limits and losing a reachable goal) stays fixed.

**Local verification:** full OMPL unit suite green — 25/25 across all eight configurators (SBL, PRM, PRMstar, LazyPRMstar, EST, BKPIECE1, KPIECE1, RRTConnect), including the two Cartesian start/goal tests and the motion validator test.

PR #771, opened 2026-08-26
TrajOptIfoptMotionPlanner: optional per-move profile-resolution trace

munich35:
A move instruction whose profile key is missing from the profile dictionary silently falls back to
the default move profile. That is the intended behaviour, but it is invisible, and a planner that
'works' with every move on the default profile is a common source of confusing results.

This adds an opt-in trace, gated once per process by the environment variable
TESSERACT_TRAJOPT_IFOPT_PROFILE_TRACE (a positive integer caps the number of lines; 0 = unlimited;
unset = disabled, costing one static bool check per move). Each line reports the move index, the
planner namespace, the requested profile key, the waypoint kind, the resolved profile's cartesian
constraint coefficients (when it is a TrajOptIfoptDefaultMoveProfile) and whether the resolution
fell back to the default profile — detected by pointer identity against a named default, which is
the only behavioural change (none: the same default is passed as before).

Function-body only, no header or class change, logs through console_bridge.

Follow-up to the discussion in #762/#763 (profile-dictionary miss diagnosis); this is the read-only trace, no behaviour change.

### DIFF
--- a/motion_planners/trajopt_ifopt/src/trajopt_ifopt_motion_planner.cpp
+++ b/motion_planners/trajopt_ifopt/src/trajopt_ifopt_motion_planner.cpp
@@ -138,10 +138,10 @@ PlannerResponse TrajOptIfoptMotionPlanner::solve(const PlannerRequest& request) 
       {
         ++trace_lines;
         const auto& wp = move_instruction.getWaypoint();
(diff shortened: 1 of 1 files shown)


Levi-Armstrong:
It is our plan to make a lot of this be explicit for the 1.0 release because as you pointed out it is unknown whether something is using the default and should it be.

Issue #768, opened 2026-08-10
Follow-up: batch bulk collision-object transform updates (one broadphase flush instead of N)


rjoomen:
**Context**

This is a planned performance follow-up related to the integer link-id migration (the LinkId/JointId/NameId PRs). The migration landed an id-typed parallel-array transform setter on the contact-manager base classes; this follow-up gives that setter batched backend implementations and its first production callers. Behavior-neutral — collision results are unchanged; only the amount of broadphase work changes.

**What this changes**

The Coal and FCL contact managers have a two-tier transform API: setting one object at a time applies the update to the broadphase immediately (a rebuild/re-balance), while the map-based setter collects all updates and applies them once. Any code that loops over the single-object setter therefore pays one broadphase rebuild per link. The base class's parallel-array setter is exactly such a loop, and so are several production call sites in the environment trajectory checks and in the TrajOpt collision evaluators — all of which run per trajectory state inside planning and optimization inner loops (and per environment state change under the continuous-monitoring path).

On Coal there's an extra cost: the per-object path can never reach the manager's full-refit fast branch, which is the branch specifically intended for the trajectory-optimization case where nearly all objects move each step.

This follow-up overrides the parallel-array setter in Coal and FCL so it collects all updates and flushes once, and rewrites the production loops to issue a single batched call per state. So a state with N active links costs one broadphase flush instead of N.

Bullet (the default backend) needs no change — its per-object update carries no batchable flush, so it does the same work either way and is unaffected.

**Expected impact**

The improvement is structural rather than marginal: on Coal and FCL, bulk transform updates on the hot planning/optimization paths go from N broadphase rebuilds to one, so the win grows with the active-link count and should be meaningful for realistic robots. Two honest caveats: users on the default Bullet backend see no change, and the in-tree transform microbenchmark moves only a single link, so it cannot show the batching win — the improvement is argued structurally and best measured on a real planning workload against Coal or FCL. New unit tests pin down that each batched override correctly refreshes the broadphase (a forgotten flush would be a correctness bug, since contactTest does not refresh it), so behavior neutrality is enforced.

**Status**

Planned, with a written implementation plan ready.

PR #774, opened 2026-09-05
Batch the OMPL validator transform updates

rjoomen:
Caller half of tesseract-robotics/tesseract#1361, which adds bulk collision object operations to the
contact manager base classes. **Merge that one first** — these call sites do not compile without the
new overloads.

### What changes

Three of the four sites — the two OMPL validators and the collision cost objective — already issued
one bulk call per state. What they also did was hand-roll the gather: copy the active links' poses
out of the state map into a thread-local vector, then pass that. They now hand the state map
straight to the `(ids, map)` overload and let the contact manager gather. Behaviour is unchanged,
including the `std::out_of_range` a missing link raises.

The one site that looped the flushing single-object setter is `checkStateInCollision` in
`motion_planners/ompl/src/utils.cpp`, reached from `applyStartStates`/`applyGoalStates` — once per
plan per IK solution, not per sampled state. That one now issues a single bulk call, so a backend
that batches performs one broadphase flush instead of one per active link.

### Performance

The measured win — ~10 us per continuous contact check on Coal, which on cheap `FIRST` checks is
+26-88% because the check itself can cost as little as 13 us — belongs to the stack rather than to
this PR. It lands in tesseract#1361; on Coal the validators' vector calls have batched since
`67d348f`. Bullet is unaffected.

### Testing

Full workspace across the five affected repositories: 1908 tests, 0 failures, 14 pre-existing skips.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Levi-Armstrong (approved):

rjoomen:
https://github.com/tesseract-robotics/tesseract_planning/issues/768

